Issue #22, opened 2019-07-17
Remove ugly hacks in client.rs 


andygrove:
I wrote the following terrible code to get the response from the future that happens when the server returns results. I feel like this would be a good place use async/await when available, or at least use channels ?

```rust
        let mut results: Arc<Mutex<Vec<proto::ExecuteResponse>>> = Arc::new(Mutex::new(vec![]));

        tokio::run(execute);

        let lock = final_result.borrow_mut().lock().unwrap();
        Ok(lock[0].clone())
```

PR #50, opened 2019-07-25
Use channel to send results back from executor to client

sd2k:
I noticed a TODO in client.rs and thought it was a nice easy change. I've tested that it works by rerunning the NYC taxis example.


sd2k:
Oh, looks like this is #22

Issue #35, opened 2019-07-20
What kubernetes constructs make most sense for the cluster?


andygrove:
In the PoC, one pod and one service is created for each executor, but how should this really work? StatefulSet? Service with ReplicaSet?

I think the distributed query planner needs the ability to target specific executors (for things like data locality later on). Let's use this issue for the initial discussion.

rrichardson:
StatefulSets have an interesting property, and that is [Stable Network Identities](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#pods-in-a-statefulset)  
This is so that handlers for partitioned datasets can be accessed directly and consistently.  I think this would be the resource for executors. 

For queries and query planning,  I think a Custom Resource Definition for Query and one for client programs would make sense. Then the query planner is an operator which reads the query planner and executes the appropriate `Jobs` that consume the executors.

andygrove:
New example uses stateful sets. Closing this issue. CRDs sound good and we should have separate issues for those.

Issue #28, opened 2019-07-19
Kubernetes URL should be configurable


andygrove:
Currently, the Kubernetes URL is hard-coded and it needs to be configurable. 

We need to think about how we want configuration to work in general. I am a fan of twelve factors [0] in general, and would encourage anyone working on this issue to read the configuration section there at least.

[0] https://12factor.net/

sd2k:
Many other clients that talk directly to Kubernetes (e.g. [helm](https://github.com/helm/helm) and [stern](https://github.com/wercker/stern)) take `--kubeconfig`, `--context` and `--namespace` arguments in their CLIs and use the standard Kubernetes config file to figure out connection details. Perhaps that model would make sense here, too?

andygrove:
Closing this as no longer relevant, since deployment now is just a yaml file so users have full control.

Issue #14, opened 2019-07-16
Remove dependency on kubectl


andygrove:
The code currently shells out to kubectl for certain operations, instead of just posting YAML directly to the k8s API.

rrichardson:
It looks like the current usage of `kubectl apply` is restricted to the creation of new resources, and not altering existing resources. 
If this is the case, it is quite a bit easier than re-implementing `kubectl apply` which  detects the appropriate resource type, and runs a 3-way patch. 

If we just need to create pods ala https://github.com/andygrove/ballista/blob/master/src/cluster.rs#L113

We should be able to just use: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#create-55

rrichardson:
I *think* that translates to https://docs.rs/k8s-openapi/0.4.0/k8s_openapi/api/core/v1/struct.CreateNamespacedPodOptional.html

andygrove:
Thanks for the info @rrichardson

andygrove:
Fixed in #55

PR #55, opened 2019-07-28
Use a StatefulSet for executor and Kubernetes API instead of kubectl

sd2k:
This commit changes a few things about how we interact with Kubernetes. Specifically:

- following some advice in #35 the executor cluster is now created as a StatefulSet with an associated headless service.
- allow existing persistent volumes to be mounted into executor pods using the `--volumes` flag of `ballista create-cluster`. This still needs work - the access mode and capacity of the PVC are hardcoded right now.
- remove dependency on kubectl (#14) and use k8s_openapi and reqwest to create/delete clusters and applications. This means we can remove the dependency on gtmpl.
- use standard methods to decide how to communicate with Kubernetes (#28): the ~/.kube/config file while outside Kubernetes, or the 'in-cluster config' (using environment variables) while inside Kubernetes - e.g. for the application pods). ~This does mean adding a git dependency for the currently unpublished `kubernetes` crate, so we should ask to get that published to crates.io!~ no longer relevant - uses the `kube` crate

A few things I noticed while writing this:

- the dev cycle is quite slow while working on the example, due to the way the Dockerfiles are laid out: making any changes to the ballista package will require a full rebuild of the example Dockerfile. Perhaps there's a better way of organising that! I'll have a think.
- error handling is going to get tricky, since there are so many possible error messages from Kubernetes alone! It might be worth looking into [snafu][] or [failure][] (although I think failure is falling out of favour?)
- it would be nice to adopt [structopt][] to handle argument parsing since we could end up with quite a lot of options!

Happy to make drastic changes to this by the way - it was partly a way for me to get stuck in! Let me know if there are parts you don't like.

[snafu]: https://docs.rs/snafu/0.4.3/snafu/
[failure]: https://github.com/rust-lang-nursery/failure
[structopt]: https://github.com/TeXitoi/structopt


andygrove:
This is looking great, thanks @sd2k !

The current example requires deploying a docker image into the cluster. I am working on the physical query plan so we can just send a plan to the cluster instead and this will speed up dev cycles.

sd2k:
Ah nice, that will be great.

Is there anything you dislike about this PR? I think it may actually be ready to merge if not. I'll create new issues to discuss the other things I mentioned.

rrichardson:
I just gave it a quick once over.  Looks good to me.  Nice work!

andygrove:
Thanks @sd2k !

PR #64, opened 2019-08-01
Use structopt for parsing CLI arguments

### LINKED ISSUE #60: Consider using structopt to define the CLI
If other tools in this space are anything to go by, we'll end up with a lot of options in the CLI, so having arg parsing handled automatically would be great.

sd2k:
This is to show how the CLI module may look if structopt is used instead of base clap.

This is based on #55 and so shouldn't be merged until that PR is (it should be simple to change this PR in case #55 isn't merged).

Fixes #60.


sd2k:
@andygrove rebased on master :+1:

PR #92, opened 2020-03-28
Kotlin: Update SQL Tokenizer for Comparison Operators

blad:
### Changes
- Expands SQLTokenizer tests by mirroring many of the tests from the SQLParer.
- Add basic tests for LiteralDoubleToken, LiteralLongToken.
- Add support for compound operators.
- Add test for compound operators.

There's more changes I'd like to make to the tokenizer to make it more robust, but thought this would be a good way to get introduced into the project.

My only question at the moment is: Is there a particular SQL implementation the Tokenizer should aim to support?


andygrove:
Hey @blad ! Sorry, I just saw this PR .. will look soon!

PR #151, opened 2020-05-19
[Rust] Remove logical plan

andygrove:
This PR removes the LogicalPlan and uses the one from DataFusion instead. Requires latest from Arrow master branch.


andygrove:
@houqp fyi

rust/ballista/src/serde/to_proto.rs
```diff
@@ -73,7 +73,7 @@ impl TryInto<protobuf::LogicalPlanNode> for LogicalPlan {
 
     fn try_into(self) -> Result<protobuf::LogicalPlanNode, Self::Error> {
         match self {
-            LogicalPlan::FileScan {
+            LogicalPlan::CsvScan {
```
andygrove, line 76: This is the main thing left to fix ... needs to handle CSV and Parquet and will require a change to the protobuf file.

houqp:
Can't wait to see this get merged :) let me know if there is anything I can help with.

rust/ballista/proto/ballista.proto
```diff
@@ -74,12 +74,13 @@ message LogicalPlanNode {
...
 
+//TODO break this out into separate CsvScanNode and ParquetScanNode
 message ScanNode {
   string path = 1;
   repeated string projection = 2;
   Schema schema = 3;
-
-  //TODO: csv and parquet specific fields
+  string file_format = 4; // parquet or csv
+  bool has_header = 5; // csv specific
```
houqp, line 83: unrelated question, why is this protobuf definition file duplicated in two places?

andygrove: Good question. I haven't found a good solution to Rust projects building with files outside of the source tree, especially when building with docker, so I took this short cut. I would like to find a better solution.

PR #153, opened 2020-05-30
[Rust] Etcd support [WIP]

andygrove:
Rust executor will register with etcd on startup if etcd urls are provided on command-line.

### DIFF
--- a/rust/ballista/src/bin/executor.rs
+++ b/rust/ballista/src/bin/executor.rs
@@ -1,4 +1,8 @@
+#![feature(async_closure)]
+
 use std::pin::Pin;
+use std::thread;
+use std::time;
 
 use ballista::datafusion::execution::context::ExecutionContext;
 use ballista::serde::decode_protobuf;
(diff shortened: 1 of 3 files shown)


rust/ballista/src/bin/executor.rs
```diff
@@ -192,18 +200,71 @@ fn to_tonic_err(e: &datafusion::error::ExecutionError) -> Status {
...
+
+            let mut client = Client::connect([&urls], None).await?;
+
+            let uuid = Uuid::new_v4();
+
+            //TODO make configurable
+            let cluster_name = "default";
+            let lease_time_seconds = 60;
+
+            let key = format!("/ballista/{}/{}", cluster_name, uuid);
```
KenSuenobu, line 241: You _may_ want to show the cluster name and UUID in logging just in case.

KenSuenobu:
Might also want to remove the unused imports in the test suite.

andygrove:
I no longer think that etc is such a good idea, so am closing this for now.

PR #154, opened 2020-05-30
Workflow improvement for GitHub Workflows

KenSuenobu:
This fixes up the build so that it builds for three platforms: Linux Ubuntu, Mac OS X, and Windows.  Windows was a little bit of a red-headed step-child since it requires the use of `vcpkg` to install the openssl library.  This increases the build time by about 2.5x, unfortunately.


andygrove:
Thanks @KenSuenobu !

PR #164, opened 2020-06-18
bump kube and k8s-openapi version

houqp:
this resolves crate version conflict caused by thrift 0.13 introduced
from latest arrow master.


rust/ballista/src/cluster.rs
```diff
@@ -73,42 +37,39 @@ impl Executor {
...
 
-    let (request, response_body) = Pod::list_namespaced_pod(
-        namespace,
-        ListNamespacedPodOptional {
-            label_selector: Some(&format!("{}={}", CLUSTER_LABEL_KEY, cluster_name)),
-            ..Default::default()
-        },
-    )?;
-    let response = execute(request, response_body)?;
+    let client = kube::client::Client::try_default().await?;
```
houqp, line 46: try_default automatically handles credential chain look up from both cluster config and kubeconfig file.

Issue #136, opened 2020-05-06
[JVM] Use a robust CSV library


andygrove:
The JVM query engine currently has some simple proprietary code for reading CSV files and we should use an established library like Apache Commons CSV or whatever Hadoop or Spark uses. We need to run some benchmarks before choosing a library to make sure we're not sacrificing on performance.

zznq:
I came across this repo that has already run some benchmarks. https://github.com/univocity/csv-parsers-comparison

It looks like OpenCSV and Apache CSV are both fairly slow on the 3 million row test file. The faster ones look to be lesser known libraries and perhaps there is a little bit of bias in the benchmarks since the fastest one is from the univocity themselves.

Not sure if this is enough to make a decision from.....

For what it's work it does look like Spark is using the `UnivocityParser`, so maybe that is the one to go with.

andygrove:
Thanks @zznq Univocity sounds good to me.

andygrove:
This was closed by https://github.com/ballista-compute/ballista/pull/165

PR #165, opened 2020-06-19
Use Univocity Parser

zznq:
# Issues:
#136 

# Overview:
Introduce a robust CSV parser, specifically the univocity parser. This parser is the one that is used by Spark, it also has some (most likely biased) benchmarks indicating that it is fast.

# Left to do for this PR:
- [x] I ignored projections to speed up the initial part of the conversion, need to add this back
- [x] add tests for special csv cases, i.e. strings with commas in them, different delimiters
- [x] add specific asserts for the tests
- [ ] infer field types instead of defaulting to strings (will create a separate PR)


jvm/datasource/src/main/kotlin/CsvDataSource.kt
```diff
@@ -44,92 +57,119 @@ class CsvDataSource(val filename: String, val schema: Schema?, private val batch
...
+        } else {
+            finalSchema
+        }
+
+        val settings = defaultSettings()
+        if(projection.isNotEmpty()) {
+            settings.selectFields(*projection.toTypedArray())
+        }
+        settings.isHeaderExtractionEnabled = hasHeaders
+        if(!hasHeaders) {
```
andygrove, line 72: nit: I would prefer to see whitespace between `if` and the expression. I guess we should decide on coding standards as as separate issue and perhaps look into code formatting tools as well.

jvm/datasource/src/main/kotlin/CsvDataSource.kt
```diff
@@ -159,10 +159,11 @@ class ReaderIterator(private val schema: Schema,
 
         do {
             val line = parser.parseNextRecord()
-            if(line != null) rows.add(line)
+            if (line != null) rows.add(line)
+            if (line != null) rows.add(line)
```
andygrove, line 163: Is it intentional that this line is repeated?

zznq: def unintentional :disappointed:

PR #167, opened 2020-06-20
add sql support in local mode & add sort support

houqp:
Build of this PR depends on https://github.com/apache/arrow/pull/7464.


rust/ballista/src/dataframe.rs
```diff
@@ -91,12 +96,52 @@ impl Configs {
     }
 }
 
+#[derive(Debug)]
+pub struct ContextSchemaProvider {
+    pub temp_tables: HashMap<String, DataFrame>,
```
houqp, line 101: This implementation is a little bit different from the one in datafusion, it allows registering arbitrary dataframe as a temporary table for access within sql query.

rust/ballista/src/dataframe.rs
```diff
@@ -91,12 +96,52 @@ impl Configs {
...
+
+#[derive(Debug)]
 pub struct Context {
-    state: Arc<ContextState>,
+    pub state: Arc<ContextState>,
+}
+
+#[derive(Debug)]
+pub struct ContextState {
+    pub schema_provider: RwLock<ContextSchemaProvider>,
```
houqp, line 139: this reference needs to be mutable due to the need to reconstruct context from any dataframe objection. it also needs to be Send due to async :(

rust/ballista/src/dataframe.rs
```diff
@@ -197,47 +271,46 @@ fn parse_settings(settings: HashMap<&str, &str>) -> HashMap<String, String> {
...
 }
 
 impl DataFrame {
     /// Create a builder from an existing plan
-    pub fn from(ctx: Arc<ContextState>, plan: &LogicalPlan) -> Self {
-        Self {
-            ctx_state: ctx,
-            plan: plan.clone(),
-        }
+    pub fn from(ctx_state: Arc<ContextState>, plan: LogicalPlan) -> Self {
```
houqp, line 284: Remove reference in plan to push clone responsibility to caller. There are cases where clone is not needed if caller intend to transfer ownership entirely.

houqp:
rebased to pull in fix from https://github.com/apache/arrow/pull/7464.

andygrove:
@houqp Could you fix the conflict?

houqp:
@andygrove rebased and builds are passing now.

andygrove:
Thanks @houqp

Issue #7, opened 2019-07-15
Implement distributed query planner for simple aggregate queries


andygrove:
Given a simple aggregate query such as the following one, Ballista should be able to create a plan to execute this query against each partition and then combine the results and run a secondary aggregate query on those.

This will also require some partition-aware meta data about files.

```sql
SELECT passenger_count, MAX(fare_amount) FROM tripdata GROUP BY passenger_count
```

andygrove:
For reference here is an example query plan from Apache Spark:

```
== Parsed Logical Plan ==
'Aggregate ['passenger_count], ['passenger_count, unresolvedalias('MIN('fare_amount), None), unresolvedalias('MAX('fare_amount), None)]
+- 'UnresolvedRelation `tripdata`

== Analyzed Logical Plan ==
passenger_count: int, min(fare_amount): double, max(fare_amount): double
Aggregate [passenger_count#3], [passenger_count#3, min(fare_amount#10) AS min(fare_amount)#44, max(fare_amount#10) AS max(fare_amount)#45]
+- SubqueryAlias `tripdata`
   +- Relation[VendorID#0,tpep_pickup_datetime#1,tpep_dropoff_datetime#2,passenger_count#3,trip_distance#4,RatecodeID#5,store_and_fwd_flag#6,PULocationID#7,DOLocationID#8,payment_type#9,fare_amount#10,extra#11,mta_tax#12,tip_amount#13,tolls_amount#14,improvement_surcharge#15,total_amount#16] csv

== Optimized Logical Plan ==
Aggregate [passenger_count#3], [passenger_count#3, min(fare_amount#10) AS min(fare_amount)#44, max(fare_amount#10) AS max(fare_amount)#45]
+- Project [passenger_count#3, fare_amount#10]
   +- Relation[VendorID#0,tpep_pickup_datetime#1,tpep_dropoff_datetime#2,passenger_count#3,trip_distance#4,RatecodeID#5,store_and_fwd_flag#6,PULocationID#7,DOLocationID#8,payment_type#9,fare_amount#10,extra#11,mta_tax#12,tip_amount#13,tolls_amount#14,improvement_surcharge#15,total_amount#16] csv

== Physical Plan ==
*(2) HashAggregate(keys=[passenger_count#3], functions=[min(fare_amount#10), max(fare_amount#10)], output=[passenger_count#3, min(fare_amount)#44, max(fare_amount)#45])
+- Exchange hashpartitioning(passenger_count#3, 200)
[...]

PR #179, opened 2020-07-05
Implement basic scheduler to create Job (DAG of stages) from a physical query plan

andygrove:
This PR introduces a basic scheduler capable of create a Job definition (DAG of stages) from a physical query plan.

Example:

```
Job 15aeb081-e515-44fa-a02c-6c5e3c59d16e has 2 stages:

Stage 0:

Stage 0 depends on stages [1].

HashAggregate: mode=Final, groupExpr=[#passenger_count], aggrExpr=[SUM(#fare_amount)]
  ShuffleReader: stage_id=1

Stage 1:

Stage 1 has no dependencies.

HashAggregate: mode=Partial, groupExpr=[#passenger_count], aggrExpr=[SUM(#fare_amount)]
  ParquetScan: "/mnt/nyctaxi/parquet/year=2019", partitions=12
```

It also lays some groundwork in defining the physical plan in protobuf format.

### DIFF
--- a/rust/ballista/src/execution/physical_plan.rs
+++ b/rust/ballista/src/execution/physical_plan.rs
@@ -39,6 +39,7 @@ use crate::datafusion::logicalplan::Expr;
 use crate::execution::filter::FilterExec;
 use crate::execution::parquet_scan::ParquetScanExec;
 use crate::execution::projection::ProjectionExec;
+use crate::execution::shuffle_reader::ShuffleReaderExec;
 use crate::execution::shuffled_hash_join::ShuffledHashJoinExec;
 
 /// Stream of columnar batches using futures
@@ -117,6 +118,8 @@ pub enum PhysicalPlan {
     ShuffledHashJoin(ShuffledHashJoinExec),
     /// Performs a shuffle that will result in the desired partitioning.
     ShuffleExchange(Rc<ShuffleExchangeExec>),
+    /// Reads results from a ShuffleExchange
+    ShuffleReader(Rc<ShuffleReaderExec>),
     /// Scans a partitioned data source
     ParquetScan(Rc<ParquetScanExec>),
 }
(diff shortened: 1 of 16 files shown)


rust/ballista/src/execution/physical_plan.rs
```diff
@@ -107,27 +108,27 @@ pub enum ColumnarValue {
...
+    Filter(Rc<FilterExec>),
     /// Hash aggregate
     HashAggregate(Rc<HashAggregateExec>),
     /// Performs a hash join of two child relations by first shuffling the data using the join keys.
     ShuffledHashJoin(ShuffledHashJoinExec),
     /// Performs a shuffle that will result in the desired partitioning.
     ShuffleExchange(Rc<ShuffleExchangeExec>),
     /// Scans a partitioned data source
-    ParquetScan(ParquetScanExec),
+    ParquetScan(Rc<ParquetScanExec>),
```
houqp, line 121: if this field rc wrap is for avoiding having to wrap rc on each match branch within `as_execution_plan`, then it's cleaner to just wrap the `Rc::new` on the whole match statement.

andygrove: Thanks. My reasoning for the Rc here is that we often have recursive code walking the plan and re-writing portions and it seemed more efficient to just clone pointers where we copy parts of the plan unmodified, rather than copy the actual struct, which could be nested fairly deep.

I'm open to changing this though once the basic functionality works.

andygrove:
Closes https://github.com/ballista-compute/ballista/issues/7

rust/ballista/src/execution/physical_plan.rs
houqp: ha, i see, in that case, better keep it as is and revisit possibility of simplicity later.

PR #239, opened 2020-07-20
Added support to filters.

### LINKED ISSUE #184: Implement execution logic for FilterExec

jorgecarleitao:
Closed #184.

Note that this currently fails the tests because we have not implemented comparison operators:

```
thread 'hash_aggregate_with_filter' panicked at 'called `Result::unwrap()` on an `Err` value: General("Unsupported binary operator GtEq")', ballista/tests/query_execution.rs:56:9
```

We need to push for that before this one.

### DIFF
--- a/rust/ballista/src/execution/operators/filter.rs
+++ b/rust/ballista/src/execution/operators/filter.rs
@@ -16,12 +16,17 @@
 
 use std::sync::Arc;
 
+use crate::arrow;
+use crate::arrow::array;
 use crate::arrow::datatypes::Schema;
 use crate::datafusion::logicalplan::Expr;
 use crate::error::{ballista_error, Result};
-use crate::execution::physical_plan::{
-    compile_expression, ColumnarBatch, ColumnarBatchIter, ColumnarBatchStream, ColumnarValue,
-    ExecutionContext, ExecutionPlan, Expression, PhysicalPlan,
+use crate::{
+    cast_array,
+    execution::physical_plan::{
+        compile_expression, ColumnarBatch, ColumnarBatchIter, ColumnarBatchStream, ColumnarValue,
+        ExecutionContext, ExecutionPlan, Expression, PhysicalPlan,
+    },
 };
 
 use async_trait::async_trait;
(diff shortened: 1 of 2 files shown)


rust/ballista/src/execution/operators/filter.rs
```diff
@@ -85,6 +97,17 @@ impl ColumnarBatchIter for FilterIter {
 }
 
 /// Filter the provided batch based on the bitmask
-fn apply_filter(_batch: &ColumnarBatch, _bitmask: &ColumnarValue) -> Result<ColumnarBatch> {
-    Err(ballista_error("filter operator is not implemented yet"))
+fn apply_filter(batch: &ColumnarBatch, bitmask: &ColumnarValue) -> Result<ColumnarBatch> {
+    let a = bitmask.to_arrow();
```
jorgecarleitao, line 101: I will simplify this later; no need for `a` and `b`.

rust/ballista/src/execution/operators/filter.rs
```diff
@@ -85,6 +97,17 @@ impl ColumnarBatchIter for FilterIter {
...
+
+    let b = a.as_any().downcast_ref::<BooleanArray>().unwrap();
```
jorgecarleitao, line 103: Remove `unwrap`

jorgecarleitao:
I should note that this code is quite similar to datafusion, and we should work on a strategy to have this declared on datafusion's side, since this code is not distributed. It seems to me that the main challenge is the sync/async.

andygrove:
Thanks @jorgecarleitao I can get comparison expressions done today most likely.

My plan is to have Ballista 0.3.0 use the official Arrow 1.0.0 release (will be released very soon) and this is why I had to duplicate code for now. Once 0.3.0 is released, I plan on creating a 0.4 branch which can use lastest DataFusion and we can start making changes there and remove code from Ballista.

Main things I need are:

- Expressions must support Debug and/or Display
- Operators must support async and I need to remove the thread per partition approach since that doesn't work for distributed

jorgecarleitao:
Got it. Full onboard; sounds good. :)

rust/ballista/src/execution/operators/filter.rs
andygrove: There is a `cast_array!` macro you can use instead.

andygrove:
@jorgecarleitao Comparison expressions are in now, at least for numeric types.

jorgecarleitao:
@andygrove , rebased on top of master. Tests now pass.

PR #249, opened 2020-07-22
Adding rudimentary cli arg parsing to example

kyprifog:
Parsing cli args simply without having to get into bringing other dependencies (like clap)


kyprifog:
Sure thing!  I was trying to avoid pulling anything new in, but if you are using it already that works

PR #268, opened 2020-07-26
Migrate to latest arrow

jorgecarleitao:
This migrates the code base to align with master from arrow.

Note that I have only validated that the tests pass, and assumed that this was a sufficient condition. However, looking at the tests themselves, this may not be sufficient.

This removes two columns from `ballista.proto`.

### DIFF
--- a/proto/ballista.proto
+++ b/proto/ballista.proto
@@ -17,9 +17,6 @@ message LogicalExprNode {
   string column_name = 10;
   bool has_column_name = 11;
 
-  uint32 column_index = 12;
-  bool has_column_index = 13;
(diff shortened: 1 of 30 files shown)


jorgecarleitao:
Also, here I followed the convention that I set in datafusion of not naming physical expressions, and instead delegate all the naming to the logical plan. In this code base this is easier to do as there is not "Merge" on the aggregations.

jorgecarleitao:
Also, this code needs [ARROW-9559](https://issues.apache.org/jira/browse/ARROW-9559) to compile.

Issue #207, opened 2020-07-13
Update Kotlin executors to reflect changes made to protobuf


andygrove:
The Kotlin executors need updating to support the `query`, `task`, and `fetch_shuffle` actions.

The `query` task should preserve the current behavior of accepting a logical plan and executing it.

The `task` action can return an error stating that the Kotlin executor doesn't yet support any physical plans. There is a follow-up PR to add support for executing Java UDFs [1] and that's where we will start implementing physical operators for the Kotlin executor to support.

The `fetch_shuffle` action can return a "shuffle not found" error for now, since the Kotlin executor doesn't yet have support for executing tasks and storing shuffle output.

andygrove:
fixed in https://github.com/ballista-compute/ballista/pull/270

PR #270, opened 2020-07-26
add Action interface to kotlin executor

kkuo:
PR for issue 207 - Update Kotlin executors to reflect changes made to protobuf

https://github.com/ballista-compute/ballista/issues/207


andygrove:
Thanks @kkuo ! This looks good overall but I don't think we should create these new classes in the `datatypes` module because we don't want datatypes to depend on the logical or physical plan ... I think that introduces a circular dependency.

Perhaps these new classes could go in the `physical-plan` module instead?

kkuo:
> Thanks @kkuo ! This looks good overall but I don't think we should create these new classes in the `datatypes` module because we don't want datatypes to depend on the logical or physical plan ... I think that introduces a circular dependency.
> 
> Perhaps these new classes could go in the `physical-plan` module instead?

sounds good. QueryAction is for logical plan which should be in the logical-plan module right? Kind of wondering where Action interface can go that will allow for both QueryAction (logical-plan) and Task (for TaskAction/ShuffleAction in the future)

andygrove:
> > Thanks @kkuo ! This looks good overall but I don't think we should create these new classes in the `datatypes` module because we don't want datatypes to depend on the logical or physical plan ... I think that introduces a circular dependency.
> > Perhaps these new classes could go in the `physical-plan` module instead?
> 
> sounds good. QueryAction is for logical plan which should be in the logical-plan module right? Kind of wondering where Action interface can go that will allow for both QueryAction (logical-plan) and Task (for physical-plan in the future)

All of the actions should be in the same module and package. The actions are really part of the physical plan because this is about how we actually execute the query in a cluster.

Issue #228, opened 2020-07-17
Remove assertions from integration test


andygrove:
The integration test [1] should use `Result` to return errors on failing tests instead of calling assertions that cause a panic.

[1] https://github.com/ballista-compute/ballista/blob/main/integration-tests/rust/src/main.rs

PR #272, opened 2020-07-27
Rewriting integration tests to use flatmap over Result

kyprifog:
Resolves https://github.com/ballista-compute/ballista/issues/228

### DIFF
--- a/integration-tests/rust/src/main.rs
+++ b/integration-tests/rust/src/main.rs
@@ -53,7 +53,7 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
-async fn execute(path: &str, host: &str, name: &&str, port: &usize) -> Result<()> {
+async fn execute(path: &str, host: &str, name: &&str, port: &usize) -> Result<String> {
     let start = Instant::now();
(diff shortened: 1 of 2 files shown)


integration-tests/rust/src/main.rs
```diff
@@ -77,38 +77,37 @@ async fn execute(path: &str, host: &str, name: &&str, port: &usize) -> Result<()
...
+        if x == y { Ok(()) } else { Err(General(format!("{:?} not equal to {:?}", x, y)))}
+    }
+
+    let mut tests = Vec::new();
+
+    tests.push(check(1, response.len()));
     let batch = &response[0];
-    assert_eq!(3, batch.num_columns());
-    assert_eq!(10, batch.num_rows());
+    tests.push(check(3, batch.num_columns()));
```
andygrove, line 88: This is looking great, thanks. Could we also pass a string message into `check`  and `check_type` so we know which check failed? For example, `check(3, batch.num_columns(), "batch has correct column count")`.

kyprifog: Sure thing!

PR #278, opened 2020-07-28
[Rust] Add workflow to build and publish docker images

jorgecarleitao:
This PR provides a mechanism to automatically build, cache and distribute rust docker images, so that they can be pulled from k8 clusters. I am hoping that this significantly improves the quality of life in contributing to this project.

The images are named based on the repository, which means that developers do not need to PR to have the image available. Furthermore, we cache all images and re-use them to reduce build time.

Below are the timings for building and publishing images on github runners:

From scratch: [21m](https://github.com/jorgecarleitao/ballista/runs/918352998?check_suite_focus=true)
With changes on cargo.toml: [17m:19s](https://github.com/jorgecarleitao/ballista/runs/918462302?check_suite_focus=true)
With changes on rust: [5m](https://github.com/jorgecarleitao/ballista/runs/918460062?check_suite_focus=true)
Without changes: [1m:30s](https://github.com/jorgecarleitao/ballista/runs/917746070?check_suite_focus=true)

Overall, IMO this is a necessity as it ensures that anyone wanting to contribute has equal opportunity to do so without the need to have minikube running on their computers, or setting up a registry and stuff.

This PR will also greatly simplify the release of docker images, as a release amounts to a new flow that re-tags an image and pushes it to docker hub. Alternatively, if we push a tag named "X.Y.Z" to the repo an image with that tag is pushed to the registry `docker.pkg.github.com` (and no docker hub is required).


jorgecarleitao:
Ahaha, this works in all cases except the one I did not tested: creating a merge request :/

README.md
```diff
@@ -99,7 +99,51 @@ along with the [blog](https://ballistacompute.org/) where news and release notes
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for information on contributing to this project.
 
+### How to build rust's docker image
 
+To build the docker image in development, use
 
+```
+docker build -f docker/rust.dockerfile ballista:latest .
```
andygrove, line 107: This is missing the `-t` flag. The full command should be `docker build -f docker/rust.dockerfile ballista:latest .`.

jorgecarleitao: good catch. Fixed.

README.md
```diff
@@ -99,7 +99,51 @@ along with the [blog](https://ballistacompute.org/) where news and release notes
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for information on contributing to this project.
 
+### How to build rust's docker image
```
andygrove, line 102: This content is great but should be in a document in the `docs` directory rather than in the top-level readme. We should link to the developer documentation (docs/README.md) from the here though.

andygrove:
This is looking good, thanks. Can we keep the `./dev/build-rust.sh` script though?

jorgecarleitao:
@andygrove , I fixed the failed build and merged the changes you recommended.

wrt to the failed build, github does not allow push images on a PR as it would allow others to inject images/packages to this repo. As such, I deactivated the push step for pull requests.

They are still available from main and the build is still available on the fork's registry and on this' registry on merge, so everything on README still applies.

Issue #261, opened 2020-07-25
[Rust] Fix regression in Rust integration test


andygrove:
I could use help debugging and fixing this.

```
Executed query against Rust executor at localhost:50051 in 5195 ms
+-----------------+-----------------------------------------------+-----------------------------------------------+
| passenger_count | MIN(Column { index: 1, name: "fare_amount" }) | MAX(Column { index: 1, name: "fare_amount" }) |
+-----------------+-----------------------------------------------+-----------------------------------------------+
| 2               | -320                                          | 492.5                                         |
| 1               | -362                                          | 623259.86                                     |
| 4               | -52                                           | 500                                           |
| 5               | -52                                           | 760                                           |
| 7               | -75                                           | 78                                            |
| 8               | 7                                             | 87                                            |
| 9               | 9                                             | 92                                            |
| 0               | -52.5                                         | 36090.3                                       |
| 6               | -52                                           | 262.5                                         |
| 3               | -100                                          | 350                                           |
+-----------------+-----------------------------------------------+-----------------------------------------------+
Schema { fields: [Field { name: "passenger_count", data_type: Int64, nullable: false, dict_id: 0, dict_is_ordered: false }, Field { name: "MIN(Column { index: 1, name: \"fare_amount\" })", data_type: Float64, nullable: true, dict_id: 0, dict_is_ordered: false }, Field { name: "MAX(Column { index: 1, name: \"fare_amount\" })", data_type: Float64, nullable: true, dict_id: 0, dict_is_ordered: false }], metadata: {} }
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Int32`,
 right: `Int64`', src/main.rs:92:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

laplab:
Hi! I would like to take a try on this. Could you please tell me where can I find test data for integration tests? It is marked by path `/mnt/nyctaxi/csv/year=2019/yellow_tripdata_2019-01.csv`

andygrove:
Thanks @laplab ... I just added some brief notes here: https://github.com/ballista-compute/ballista/blob/main/docs/nyctaxi.md

laplab:
I propose we close that issue since the problem is resolved

PR #286, opened 2020-07-29
Respect schema and projection parameters from client

laplab:
The original problem of failing integration test #261 was a symptom to several issues.

Integration test failed because type of `passenger_count` column was `Int64` instead of `Int32` which was provided in schema. This happened because CSV schema (along with some other options) was not actually serialized to protobuf before sending query to the executor. The executor infered schema instead and this is the source of `Int64` type in the result.

The second issue is that even though the projection parameter is optional in the plan description, it is serialized as repeated filed in protobuf message `ScanNode`. Repeated fields cannot be optional, so empty array was sent to the executor as the projection. This resulted in empty result from CSV scan because empty projection simply means "the user does not want any fields from me".

The last issue was related to the serialization of CSV scan execution node. `CsvScanExec` inside its `try_new` method filters schema so that only projected columns remain. This projected schema is then serialized into protobuf, but the original projection field remains. This causes index error when `CsvScanExec` is restored from protobuf because projected schema simply does not contain fields mentioned in projection field. I made a quick fix for this and serialized original schema into protobuf.

Unfortunately, I was not able to debug one more problem. Even though code builds successfully on my local machine, for some reason build inside docker does not see new protobuf message I added. I have tried to rebuild dependencies docker image, but still no luck :(

### DIFF
--- a/rust/ballista/src/serde/from_proto.rs
+++ b/rust/ballista/src/serde/from_proto.rs
@@ -53,6 +53,7 @@ macro_rules! convert_box_required {
         }
     }};
 }
+
 impl TryInto<LogicalPlan> for &protobuf::LogicalPlanNode {
     type Error = BallistaError;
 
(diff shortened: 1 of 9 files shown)


rust/ballista/src/serde/from_proto.rs
```diff
@@ -98,21 +99,27 @@ impl TryInto<LogicalPlan> for &protobuf::LogicalPlanNode {
...
+
+                    let mut projection = None;
+                    if let Some(column_names) = &scan.projection {
+                        let column_indices = column_names
+                            .columns
+                            .iter()
+                            .map(|name| schema.index_of(name))
+                            .collect::<Result<Vec<usize>, _>>()?;
+                        projection = Some(column_indices);
+                    }
```
laplab, line 117: I would appreciate some help with this code. I cannot do `scan.projection.as_ref().map(...)` here because `index_of` can fail and we need to propagate error with `?` operator

andygrove:
@laplab Thanks! Great findings. I can explan the docker issue .. there are two copies of the protobuf def so you need to apply the same changes to the one in the top level protobuf directory. When building the Docker image, this one gets copies into the Rust project. It's a safety mechanism to make sure the two stay in sync.

laplab:
Just to ensure that everything is okay with updates in protobufs in JVM and Spark executors I would like to build docker images for all three executors and run integration test.
Unfortunately, docker build for JVM and Spark executors fails because of this line https://github.com/ballista-compute/ballista/blob/main/docker/jvm-executor.dockerfile#L3
@andygrove Could you please help me to figure out where this `build/distributions` folder is coming from?

andygrove:
Yes, these projects need to built first with Gradle. There is a build-all.sh in the top level dev folder. It this is too much hassle, I can pull your branch and test before merging.

laplab:
I have managed to rebuild all three images. Integration test successfully passed: https://pastebin.com/GCpjAUJF
@andygrove Do you have any suggestions or should we merge it?

PR #292, opened 2020-07-30
Migrate to latest arrow (with sqlparser:0.6.1)

jorgecarleitao:


jorgecarleitao:
I can see that because we only test PRs to main: the workflow does not run for `branch-0.4`. I ran `cd rust/ballista && cargo test` locally and they passed.

PR #295, opened 2020-07-30
Small improvement to taxi data fetching script

kyprifog:
Found this easier to work with

### DIFF
--- a/dev/download-nyctaxi-files.sh
+++ b/dev/download-nyctaxi-files.sh
@@ -1,13 +1,12 @@
 #!/bin/bash
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-01.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-02.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-03.csv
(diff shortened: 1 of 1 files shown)


dev/download-nyctaxi-files.sh
```diff
@@ -1,13 +1,12 @@
...
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-12.csv
\ No newline at end of file
+
+if [ -z "$1" ] || [ -z "$2" ]
+then
+  echo "Download files <<Color: (yellow, green, fhv)>> <<Years: 2019,2020, etc (2009-2020)>>"
+else
+  for i in $(echo $2 | sed "s/,/ /g")
+  do
+    aws s3 cp --recursive --exclude="*" --include="$1_tripdata_$i*.csv" s3://nyc-tlc/trip\ data/ /mnt/nyctaxi/$1/year=$i/
```
jorgecarleitao, line 9: Doesn't this assume having `aws` cli installed?

kyprifog: Yeah, i thought since the data is being pulled from s3 that would be a safe assumption?   I can modify to use wget like the other though

jorgecarleitao: IMO wget is much more prevalent than aws cli and thus it is much less likely that the user would need to install it just to download some public files.

Issue #215, opened 2020-07-14
Review and update k8s deployment instructions


andygrove:
The content in https://github.com/ballista-compute/ballista/tree/main/kubernetes is almost certainly out of date and needs reviewing / testing.

PR #300, opened 2020-07-31
Refactored kubernetes guide to support a remote cluster

jorgecarleitao:
The aim of this PR is to have a simple configuration that new users can use to deploy Ballista on a generic kubernetes cluster.

There are two main differences between a generic k8 and minikube from the current deployment's perspective:

* docker images must be pulled from somewhere (and not just built)
* files must be dumped in persistent volumes (and not just mounted)

This commit:

* Adds a yaml with a job to populate the data to the persistent volume, so that we do not rely on local mounts
* Adds a yaml configuration that:
    * has a `imagePullSecrets` to kubernetes be able to pull images from github.
    * uses the image `docker.pkg.github.com/ballista-compute/ballista/ballista-rust:main`, so that when this code is pulled from `main`, that is also the image that is deployed on kubernetes
    * significantly reduces the requirements for the cluster: 2 replicas instead of 12, no memory or cpu needs
* generalizes the instructions to deploy ballista on any kubernetes cluster, thus allowing the user to run it out-of-the box on a managed cluster.
* removes most code on `kubernetes/README.md`, since we have it on the repository as yaml files. This reduces the maintenance cost of maintaining two copies of the files (that had already drifted).
* reduces verbs on the ClusterRole to only list pods, since the other verbs are not needed.

At some point we should have an Helm Chart, as we should ideally have a chart that allows the different configurations.

In this journey, the largest pain point was the time it takes to download of the data. I think that we should have a smaller dataset for new users play with.

### DIFF
--- a/dev/download-nyctaxi-files.sh
+++ b/dev/download-nyctaxi-files.sh
@@ -1,13 +1,19 @@
 #!/bin/bash
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-01.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-02.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-03.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-04.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-05.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-06.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-07.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-08.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-09.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-10.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-11.csv
-wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-12.csv
+ARG=
+if [ -z "${1}" ];
+then
+ARG="-P $1"
+fi
+
+wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-01.csv $ARG
+wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-02.csv $ARG
+wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-03.csv $ARG
+wget https://s3.amazonaws.com/nyc-tlc/trip+data/yellow_tripdata_2019-04.csv $ARG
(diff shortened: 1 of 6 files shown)


andygrove:
This is looking great! We could potentially just download one of the data files and use some bash magic to partition it to break it down into smaller files, as described in this SO post:

https://stackoverflow.com/questions/1411713/how-to-split-a-file-and-keep-the-first-line-in-each-of-the-pieces

jorgecarleitao:
@andygrove , to understand, the underlying goal of your idea would be to have multiple executors executing the same query, and for that we need more than one partition, and for that we (currently) need more than one file.

andygrove:
That's correct. The distribution model is currently simplistic and one file = one partition = one task that can be run in parallel. Later, we will add support for processing sections (chunks) of a file in parallel but we don't have that yet.

andygrove:
Closes https://github.com/ballista-compute/ballista/issues/215

jorgecarleitao:
Got it. I will then download a single file, split it in 4 or so and re-deploy this on my k8 cluster again to validate everything.

jorgecarleitao:
I changed the script to download a single file and split it in 10 parts (csv):

```
Connecting to s3.amazonaws.com (52.217.12.206:443)
saving to '/mnt/yellow_tripdata_2019-01.csv'
yellow_tripdata_2019   3% |*                               | 22.7M  0:00:27 ETA
yellow_tripdata_2019  10% |***                             | 66.0M  0:00:17 ETA
yellow_tripdata_2019  16% |*****                           |  110M  0:00:14 ETA
yellow_tripdata_2019  23% |*******                         |  154M  0:00:12 ETA
yellow_tripdata_2019  30% |*********                       |  199M  0:00:11 ETA
yellow_tripdata_2019  37% |***********                     |  242M  0:00:10 ETA
yellow_tripdata_2019  43% |**************                  |  287M  0:00:08 ETA
yellow_tripdata_2019  50% |****************                |  331M  0:00:07 ETA
yellow_tripdata_2019  57% |******************              |  376M  0:00:06 ETA
yellow_tripdata_2019  64% |********************            |  420M  0:00:05 ETA
yellow_tripdata_2019  71% |**********************          |  465M  0:00:04 ETA
yellow_tripdata_2019  77% |************************        |  509M  0:00:03 ETA
yellow_tripdata_2019  84% |***************************     |  553M  0:00:02 ETA
yellow_tripdata_2019  91% |*****************************   |  597M  0:00:01 ETA
yellow_tripdata_2019  97% |******************************* |  642M  0:00:00 ETA
yellow_tripdata_2019 100% |********************************|  655M  0:00:00 ETA
'/mnt/yellow_tripdata_2019-01.csv' saved
Connecting to raw.githubusercontent.com (199.232.56.133:443)
writing to stdout
splitting csv by 766779 lines to /mnt/nztaxi.csv...
-                    100% |********************************|   422  0:00:00 ETA
written to stdout
done.
```

The files are `/mnt/nztaxi.csv/part-aa.csv`, `/mnt/nztaxi.csv/part-ab.csv`, ...

The only part that is still not documented in the guide is the execution of the client.

Essentially, the issue is that the client needs to read from `/mnt` to infer the file's schema (bug?), or we need to specify the schema manually. The latter requires hard-coding the schema in the client, which is not very portable.

So, currently the guide would go along the lines of "change this example to add this schema, compile it, and run it against 50051". IMO as a user I should not have to compile a rust program to play with it.

Another alternative is to offer an SQL-based client e.g. as a docker image, which the user would run as 

```
docker run --rm -t ... "CREATE EXTERNAL ... taxis; SELECT COUNT(*) FROM taxis"
```

For now I would say that this PR is complete in the sense that the guide describes how to deploy ballista on a cluster, just not how to use.

Issue #95, opened 2020-04-04
[JVM] SQL Parser: Add support for ORDER BY


andygrove:

andygrove:
Closing this since JVM query engine is not the current focus

PR #313, opened 2020-08-04
parse order by expression

meijies:
https://github.com/ballista-compute/ballista/issues/95

### DIFF
--- a/jvm/sql/src/main/kotlin/Expressions.kt
+++ b/jvm/sql/src/main/kotlin/Expressions.kt
@@ -58,5 +58,6 @@ data class SqlSelect(
     val projection: List<SqlExpr>,
     val selection: SqlExpr?,
     val groupBy: List<SqlExpr>,
+    val orderBy: List<SqlExpr>,
     val tableName: String
 ) : SqlRelation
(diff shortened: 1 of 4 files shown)


jvm/sql/src/main/kotlin/SqlParser.kt
```diff
@@ -138,7 +138,13 @@ class SqlParser(val tokens: TokenStream) : PrattParser {
         groupBy = parseExprList()
       }
 
-      return SqlSelect(projection, filterExpr, groupBy, table.id)
+      // parse optional ORDER BY clause
+      var orderBy: List<SqlExpr> = listOf()
+      if (tokens.consumeKeywords(listOf("ORDER", "BY"))) {
+        orderBy = parseExprList()
```
andygrove, line 144: This is a good start, but order by expressions can also have optional `ASC` or `DESC` qualifiers. For example, `ORDER BY start_date DESC, salary ASC`.

meijies: Optional `ASC` or `DESC` has supported now. You can review again

jvm/sql/src/main/kotlin/Expressions.kt
```diff
@@ -52,11 +52,14 @@ data class SqlAlias(val expr: SqlExpr, val alias: SqlIdentifier) : SqlExpr
 
 data class SqlCast(val expr: SqlExpr, val dataType: SqlIdentifier) : SqlExpr
 
+data class SqlSort(val expr: SqlExpr, val asc: String) : SqlExpr
```
andygrove, line 55: I think `asc` should be a boolean since there can only be two values - ascending or descending.

andygrove:
Thanks @meijies this looks great! If you could just change `asc` to a boolean, I'll go ahead and merge this.

meijies:
@andygrove the change has been done, you can review again

PR #319, opened 2020-08-08
[Rust] Remove async from ColumnBatchIter

andygrove:
Remove `async` from `ColumnBatchIter` so we can remove the separate threads for parquet scans.

### DIFF
--- a/proto/ballista.proto
+++ b/proto/ballista.proto
@@ -144,7 +144,6 @@ message ScanExecNode {
   string file_format = 4; // parquet or csv
   bool has_header = 5; // csv specific
   uint32 batch_size = 6;
-  uint32 queue_size = 7; // parquet specific
 
   // partition filenames
   repeated string filename = 8;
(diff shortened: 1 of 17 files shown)


jorgecarleitao:
Nice!

andygrove:
@jorgecarleitao It actually makes performance worse though ... not what I expected at all

jorgecarleitao:
That is odd.

If switching threads within a batch is worth, then there are some non-trivial waits within a batch that makes a thread to idle (and in async switches). That is unexpected under the hypothesis that we are CPU-bounded within a batch (as we do not communicate). Does this make sense?

The only lock operation I remember on the top of my head is on the aggregation, on that map of aggregator sets...

andygrove:
I think I understand better now and I have run more tests.

With the async approach, we have a thread reading batches from same parquet at the same time that we have threads processing those batches, so really we were running with double the threads.

So running with 12 executor threads with async really meant we had 24 threads, so I can't compare the results directly.

12 (really 24) threads with async took 44 seconds.
24 threads no async takes 32 seconds.

This is better after all :-)

PR #323, opened 2020-08-14
Branch 0.4 rebased

jorgecarleitao:
@andygrove , this is the rebase. The test sometimes fail but it is a trivial issue: the order of the rows is not set by the plan, and thus sometimes the rows are in a different order.

I can fix it by end of day UTC, but in case you are waiting for this, I am PRing already.


rust/ballista/src/execution/operators/projection.rs
```diff
@@ -40,16 +41,24 @@ pub struct ProjectionExec {
...
+        let fields: Result<Vec<_>> = exprs
             .iter()
-            .map(|e| e.to_schema_field(&input_schema))
+            .map(|expr| {
+                let name = expr.name(&input_schema)?;
+                let name = name.as_str();
+                Ok(Field::new(
+                    name,
+                    expr.get_type(&input_schema)?,
+                    input_schema.field_with_name(name)?.is_nullable(),
```
jorgecarleitao, line 53: this is an issue: nunable is being driven by the logical plan. to think about

rust/ballista/src/execution/physical_plan.rs
```diff
@@ -195,52 +177,53 @@ pub type MaybeColumnarBatch = Result<Option<ColumnarBatch>>;
...
     }
 
-    pub fn from_values_and_schema(values: &[ColumnarValue], schema: Arc<Schema>) -> Self {
-        Self {
-            schema,
-            columns: values.to_vec(),
-        }
-    }
-
-    pub fn from_values_infer_schema(values: &[ColumnarValue]) -> Self {
```
jorgecarleitao, line 221: we no longer support infer schema from values with this. I did not see the need for it

rust/ballista/src/serde/to_proto.rs
```diff
@@ -225,7 +225,7 @@ impl TryInto<protobuf::LogicalExprNode> for &Expr {
             Expr::Column(name) => {
                 let mut expr = empty_expr_node();
                 expr.has_column_name = true;
-                expr.column_name = name.to_owned();
```
jorgecarleitao, line 228: not sure this should be changed

jorgecarleitao:
@andygrove , this is now ready to review. I am not very familiar with the proto stuff, so that part may be a mistake.

andygrove:
Awesome. Thanks @jorgecarleitao I will review over the weekend and get this merged.

PR #328, opened 2020-09-03
specify ref for arrow dependency git repository

Dyqer:
specify ref for arrow dependency git repository to avoid cargo build failed


andygrove (approved):
LGTM. Thanks @Dyqer

PR #337, opened 2021-01-09
Fix test

edrevo:
A minor fix in one of the tests: the sort method returns Unit, so the assert would always succeed.


andygrove (approved):
LGTM. Thanks @edrevo

PR #345, opened 2021-01-11
Add project goals to CONTRIBUTING

edrevo:
Text is copy&paste from https://andygrove.io/2021/01/ballista-2021/, but added mention to Etcd since I believe K8s isn't the only resource manager that should be supported in the long term.


andygrove (approved):
Thanks @edrevo

PR #357, opened 2021-01-13
Implemented Serde for SortNode

pjmore:
Hi, I took a crack at implementing Serde for LogicalPlan::Sort. I assumed that ballista.proto needed to be changed, I'm not sure about the protobuf field numbering as there is the jump from one byte to 2 when exceeding 15 fields and I wasn't sure if the lower numbers were being reserved common node types. If this looks alright I'll take a crack at the other node types.

### DIFF
--- a/rust/ballista/proto/ballista.proto
+++ b/rust/ballista/proto/ballista.proto
@@ -86,6 +86,7 @@ message LogicalPlanNode {
   LimitNode limit = 22;
   AggregateNode aggregate = 23;
   JoinNode join = 24;
+  SortNode sort = 25;
 }
 
 message ProjectionColumns {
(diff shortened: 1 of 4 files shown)


andygrove:
Thanks @pjmore this is looking good. The `Sort` operator is a bit special though because its expressions are of type `Expr::Sort` which has some extra information that affects sorting:

```rust
    /// A sort expression, that can be used to sort values.
    Sort {
        /// The expression to sort on
        expr: Box<Expr>,
        /// The direction of the sort
        asc: bool,
        /// Whether to put Nulls before all other data values
        nulls_first: bool,
    },
```

If a Sort logical plan has expressions that are not of this type then it should be considered an error.

andygrove:
Actually, please disregard my comment.

andygrove (approved):
Thanks @pjmore this LGTM. When we implement the serialization for logical expressions we can take care of serializing `Expr::SortExpr` correctly.

PR #359, opened 2021-01-14
Logical plan node serde implementation

pjmore:
Implemented serde for the LogicalPlan nodes repartition, create external table, empty relation, and explain. I'm a little unsure about the create external table implementation because as far as I could tell none of the other node types  serialize their schema and I couldn't find a way around serializing it for this node.  

Some smaller changes made:

1.  Changed [build.rs](rust/ballista/build.rs) so it is only run when [ballista.proto](rust/ballista/proto/ballista.proto) is changed. I also changed from a panic to returning a result as I was finding the stack traces were  noise when changing the protobuf definition.

2.  Moved logical plan roundtrip tests to [src/serde/logical_plan/mod.rs](rust/ballista/src/serde/logical_plan/mod.rs).

### DIFF
--- a/rust/ballista/proto/ballista.proto
+++ b/rust/ballista/proto/ballista.proto
@@ -86,6 +86,7 @@ message LogicalPlanNode {
   LimitNode limit = 22;
   AggregateNode aggregate = 23;
   JoinNode join = 24;
+  SortNode sort = 25;
 }
 
 message ProjectionColumns {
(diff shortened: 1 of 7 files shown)


rust/ballista/proto/ballista.proto
```diff
@@ -124,6 +128,45 @@ message SortNode{
     repeated LogicalExprNode expr = 1;
 }
 
+message RepartitionNode{
+    oneof partition_method{
```
andygrove, line 132: I didn't know you could do this in protobuf - I think there are other places where we could use this

andygrove (approved):
This is great. Thank you @pjmore  !

PR #367, opened 2021-01-17
Added rust code formatting check to CI. Closes #364 

### LINKED ISSUE #364: Add cargo fmt check to CI
We want CI to fail on PRs if `cargo fmt` has not been run

pjmore:
Added a rust code formatting check to CI. 
Added job dependency so that tests run only after lints and formatting checks pass. I wasn't sure what the Docker job does so i didn't change it.
Added rule to only run rust CI when a file in rust/ is changed.

It might be worth it to check a rustfmt.toml file into the repo just to make sure that there aren't any formatting inconsistencies between contributor configurations.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,12 +1,7 @@
 name: Rust
-# Runs rust ci only when files in rust directory have been changed
-on: 
-  push:
-    paths: 
-      - 'rust/**'
-  pull_request:
-    paths:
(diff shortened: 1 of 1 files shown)


pjmore:
It looks like the directory filter I tried to use caused the rust checks to not run so I reverted it back to what it was before.

andygrove (approved):
Awesome. Thanks @pjmore I appreciate it.

PR #379, opened 2021-01-17
Implement serde code for logical expressions "Not, IsNull, IsNotNull"

### LINKED ISSUE #368: Implement serde code for logical expressions "Not, IsNull, IsNotNull"
The module at  `rust/src/serde/logical_plan` provides serde code to convert DataFusion logical plans to/from protobuf format.

It does not currently implement serde for the expression(s) mentioned in the title  of this issue. The goal for this issue is to implement the serde code  (both to and from protobuf) for these expressions, and ensure that there is a unit test demonstrating round-trip serde for a logical plan using these expressions.

It is possible that there is existing code in `rust/ballista-old/src/serde` that can be ported over. If not, it is usually possible to figure out  an approach by looking at the serde code for the expressions that are  implemented.

It may be necessary to add new expressions to the .proto file as well.

edrevo:
Still wrapping my head around DataFusion and Ballista, so fire away with any feedback you have!

Fixes #368

### DIFF
--- a/rust/ballista/proto/ballista.proto
+++ b/rust/ballista/proto/ballista.proto
@@ -12,41 +12,51 @@ option java_outer_classname = "BallistaProto";
 
 // logical expressions
 message LogicalExprNode {
+  oneof ExprType {
+    // column references
+    string column_name = 10;
 
-  // column references
-  string column_name = 10;
-  bool has_column_name = 11;
+    // alias
+    AliasNode alias = 14;
 
-  // alias
-  AliasNode alias = 14;
+    // literals
+    string literal_string = 20;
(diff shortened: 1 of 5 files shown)


rust/ballista/proto/ballista.proto
```diff
@@ -12,41 +12,51 @@ option java_outer_classname = "BallistaProto";
 
 // logical expressions
 message LogicalExprNode {
+  oneof ExprType {
```
andygrove, line 15: This is a nice improvement

andygrove (approved):
LGTM! Thanks @edrevo

PR #380, opened 2021-01-18
Implement serdes for LogicalExprNode: Between, Case, Cast, Sort

### LINKED ISSUE #369: Implement serde code for logical expression "Between"
The module at  `rust/src/serde/logical_plan` provides serde code to convert DataFusion logical plans to/from protobuf format.

It does not currently implement serde for the expression(s) mentioned in the title  of this issue. The goal for this issue is to implement the serde code  (both to and from protobuf) for these expressions, and ensure that there is a unit test demonstrating round-trip serde for a logical plan using these expressions.

It is possible that there is existing code in `rust/ballista-old/src/serde` that can be ported over. If not, it is usually possible to figure out  an approach by looking at the serde code for the expressions that are  implemented.

It may be necessary to add new expressions to the .proto file as well.

bjcohen:
I wasn't sure how to deal with the `DataType` <> `ArrowType` conversion for the `Cast` operator, since it looks like
(per https://github.com/apache/arrow/blob/master/rust/arrow/src/compute/kernels/cast.rs#L51) there are many
types of casts that are possible that weren't representable with the currently serializable types. I didn't want to change the
proto `ArrowType` in this PR, so I only added types to `from_proto_arrow_type`/`to_proto_arrow_type` that didn't have size
params.

Fixes #369, #370, #371, #372

### DIFF
--- a/rust/ballista/proto/ballista.proto
+++ b/rust/ballista/proto/ballista.proto
@@ -44,6 +44,11 @@ message LogicalExprNode {
     IsNull is_null_expr = 60;
     IsNotNull is_not_null_expr = 61;
     Not not_expr = 62;
+
+    BetweenNode between = 70;
+    CaseNode case_ = 71;
+    CastNode cast = 72;
+    SortExprNode sort = 73;
   }
 }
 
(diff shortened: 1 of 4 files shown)


andygrove (approved):
LGTM! Thanks @bjcohen

Issue #374, opened 2021-01-17
Update Kotlin code for latest protobuf changes


andygrove:
The Rust implementation currently uses a separate copy of the ballista.proto file (located at `rust/ballista/proto/ballista.proto`) which is now more up to date than the top level `proto/ballista.proto` that the JVM implementation uses.

This issue is for updating the top level proto file and fixing any Kotlin compilation issues that this causes.

PR #387, opened 2021-01-18
Updated JVM protobuf definitions

pjmore:
Migrated the JVM implementation to the new protobuf schema and sketched out a logical plan join node. Just a heads up, I'm not particularly familiar with Kotlin,  is there a rustfmt equivalent for  Kotlin?
#374


andygrove (approved):
@pjmore This is great, thank you. We use spotless for formatting. You should be able to run something like `./gradlew spotlessApply` to automatically format the code.

andygrove (approved):
Thanks @pjmore this is a huge help. I suppose this is going to be an ongoing issue keeping the two copies of this file in sync.

PR #389, opened 2021-01-18
Implement serdes for LogicalExprNode: Negative, InList, Wildcard

### LINKED ISSUE #388: Implement serde code for logical expressions: Negative, InList, Wildcard
Description copied from https://github.com/ballista-compute/ballista/issues/372:

> The module at `rust/src/serde/logical_plan` provides serde code to convert DataFusion logical plans to/from protobuf format.
> 
> It does not currently implement serde for the expression(s) mentioned in the title of this issue. The goal for this issue is to implement the serde code (both to and from protobuf) for these expressions, and ensure that there is a unit test demonstrating round-trip serde for a logical plan using these expressions.
> 
> It is possible that there is existing code in `rust/ballista-old/src/serde` that can be ported over. If not, it is usually possible to figure out an approach by looking at the serde code for the expressions that are implemented.
> 
> It may be necessary to add new expressions to the .proto file as well.

bjcohen:
Fixes #388

### DIFF
--- a/rust/ballista/proto/ballista.proto
+++ b/rust/ballista/proto/ballista.proto
@@ -49,6 +49,9 @@ message LogicalExprNode {
     CaseNode case_ = 71;
     CastNode cast = 72;
     SortExprNode sort = 73;
+    NegativeNode negative = 74;
+    InListNode in_list = 75;
+    bool wildcard = 76;
   }
 }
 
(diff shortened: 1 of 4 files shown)


andygrove (approved):
LGTM. Thanks @bjcohen

PR #391, opened 2021-01-20
Use async sleep to avoid blocking thread

edrevo:
Since this method is called through `tokio::spawn`, using `thread::sleep` isn't a great idea since it will completely block one of tokio's threads in its threadpool. By using tokio's version of sleep (called delay_for in tokio 0.2 but called sleep in tokio 1.0), the thread can handle other tasks while the etcd loop is idle.


andygrove (approved):
LGTM. This is good to know.

PR #397, opened 2021-01-22
add integration-tests to rust github actions

jameswinegar:
currently in rust.yml because the integration test depends on docker images being created.


andygrove (approved):
Thanks @jameswinegar I will follow up with another PR today to enable the tests hopefully

PR #400, opened 2021-01-22
Update arrow to 3.0.0 release

edrevo:
Update arrow to 3.0.0 release commit


andygrove (approved):
LGTM. We can update to the actual release once it is on crates.io (in the next few days I expect).

PR #402, opened 2021-01-23
Simple fix to make populate-data atomic

morpheyesh:
Fix so pod restarts on minikube will pull the data in again and will not fail to download data after a first failure.


andygrove (approved):
Thanks @morpheyesh !

PR #403, opened 2021-01-23
Fix table provider serde bugs

andygrove:
This follows on from https://github.com/ballista-compute/ballista/pull/401 and fixes some serde issues I found when trying to get TPC-H working.

Closes https://github.com/ballista-compute/ballista/issues/405


rust/ballista/src/serde/logical_plan/to_proto.rs
```diff
@@ -36,18 +37,37 @@ impl TryInto<protobuf::LogicalPlanNode> for &LogicalPlan {
...
-                    .collect();
-                let projection = Some(protobuf::ProjectionColumns { columns });
+
+                // unwrap the DFTableAdapter to get to the real TableProvider
+                let source = if let Some(adapter) = source.as_any().downcast_ref::<DFTableAdapter>()
+                {
+                    match &adapter.logical_plan {
+                        LogicalPlan::TableScan { source, .. } => Ok(source.as_any()),
+                        _ => Err(BallistaError::General(
+                            "Invalid LogicalPlan::TableScan".to_owned(),
```
edrevo, line 52: if the only valid LogicalPlan inside a DFTableAdapter is a TableScan, why can the DFTableAdapter hold other types of LogicalPlans? are you anticipating the need for more kinds of plans?

rust/ballista/src/context.rs
```diff
@@ -121,18 +121,32 @@ impl BallistaContext {
...
+/// This ugly adapter is needed because we use DataFusion's logical plan when building queries
+/// and when we register tables with DataFusion's `ExecutionContext` we need to provide a
+/// TableProvider which is effectively a wrapper around a physical plan. We need to be able to
+/// register tables so that we can create logical plans from SQL statements that reference these
+/// tables.
+pub(crate) struct DFTableAdapter {
+    /// DataFusion logical plan
+    pub logical_plan: LogicalPlan,
+    /// DataFusion execution plan
     plan: Arc<dyn ExecutionPlan>,
```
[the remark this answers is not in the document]
andygrove: The physical plan is not actually needed. I removed it and added some documentation in a follow-up PR (https://github.com/ballista-compute/ballista/pull/411).

The bug with the projection wsan't related to this code but to the serde code in to_proto where it was building a projection of all columns and ignoring the projection in the logical plan.

rust/ballista/src/serde/logical_plan/to_proto.rs
andygrove: That's a good point. It can only be a table scan, but as far as I know, it is not possible to declare a struct variable with the type `LogicalPlan::TableScan`.

edrevo: You're right. There's an unimplemented Rust RFC for that https://github.com/rust-lang/rfcs/pull/2593

PR #415, opened 2021-01-23
Refactor LogicalPlan protobuf

edrevo:
This is an opinionated change, but since it was well received in the LogicalExprNode, I thought I'd do the same for LogicalPlans too.


andygrove (approved):
LGTM. Thanks @edrevo

PR #416, opened 2021-01-23
Implement Parquet projection serde

edrevo:


rust/ballista/src/serde/logical_plan/from_proto.rs
```diff
@@ -113,7 +113,29 @@ impl TryInto<LogicalPlan> for &protobuf::LogicalPlanNode {
...
+                                    "Parquet projection contains column name that is not present in schema. Column name: {}. Schema columns: {:?}",
+                                    col_name,
+                                    column_names))
+                            })
+                        )
+                        .collect();
+                    Some(r?)
+                }
+            };
+            LogicalPlanBuilder::scan_parquet(&scan.path, projection, 24)? //TODO concurrency
```
edrevo, line 138: Note that concurrency can't be serialized since Datafusion doesn't provide an accessor for it. Should I open a bug in Datafusion/Arrow?

andygrove: Concurreny in the plan is left over from when DataFusion spawned threads and before it moved to Tokio, so I see this going away at some point.

andygrove:
@edrevo This needs a rebase now. Thanks.

edrevo:
Rebased

PR #417, opened 2021-01-23
Remove CountDistinct aggregate function

edrevo:
As far as I can tell, there's no CountDistinct aggregate function in datafusion. Maybe this is something from the past?


andygrove (approved):
LGTM

PR #423, opened 2021-01-24
Add SIMD, snmalloc options, document lto and update DataFusion

Dandandan:
* Add snmalloc option
* Add SIMD option
* Upgrade versions
* Document usage of LTO

This should give some perf improvements on the TCPH queries.

### DIFF
--- a/benchmarks/tpch/Cargo.toml
+++ b/benchmarks/tpch/Cargo.toml
@@ -7,10 +7,10 @@ edition = "2018"
 [dependencies]
 ballista = { path="../../rust/ballista" }
 
-arrow = { git = "https://github.com/apache/arrow", rev="10f4ada" }
-arrow-flight = { git = "https://github.com/apache/arrow", rev="10f4ada" }
-datafusion = { git = "https://github.com/apache/arrow", rev="10f4ada" }
-parquet = { git = "https://github.com/apache/arrow", rev="10f4ada" }
+arrow = { git = "https://github.com/apache/arrow", rev="d613aa6" }
+arrow-flight = { git = "https://github.com/apache/arrow", rev="d613aa6" }
+datafusion = { git = "https://github.com/apache/arrow", rev="d613aa6" }
+parquet = { git = "https://github.com/apache/arrow", rev="d613aa6" }
 
 env_logger = "0.8"
 tokio = { version = "0.2", features = ["full"] }
--- a/benchmarks/tpch/Cargo.toml
+++ b/benchmarks/tpch/Cargo.toml
@@ -7,10 +7,10 @@ edition = "2018"
 [dependencies]
 ballista = { path="../../rust/ballista" }
 
-arrow = { git = "https://github.com/apache/arrow", rev="d613aa6" }
-arrow-flight = { git = "https://github.com/apache/arrow", rev="d613aa6" }
-datafusion = { git = "https://github.com/apache/arrow", rev="d613aa6" }
-parquet = { git = "https://github.com/apache/arrow", rev="d613aa6" }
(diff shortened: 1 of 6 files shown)


benchmarks/tpch/Cargo.toml
```diff
@@ -7,10 +7,10 @@ edition = "2018"
 [dependencies]
 ballista = { path="../../rust/ballista" }
 
-arrow = { git = "https://github.com/apache/arrow", rev="d613aa6" }
-arrow-flight = { git = "https://github.com/apache/arrow", rev="d613aa6" }
-datafusion = { git = "https://github.com/apache/arrow", rev="d613aa6" }
-parquet = { git = "https://github.com/apache/arrow", rev="d613aa6" }
+arrow = { git = "https://github.com/apache/arrow", rev="10f4ada" }
```
andygrove, line 10: I assume that this moves us beyond the 3.0.0 release?

Dandandan: Ah yes, this is just the latest revision. Do you want to keep it on 3.0?

andygrove: I haven't really made a decision on what the strategy should be, but it would be nice to be able to release to crates.io against 3.0.0 ... that said, it is likely that we will need new DataFusion features at some point soon.

Perhaps we can start out by staying on 3.0.0 (release should be this week I hope) and then only upgrade if there is something we really need?

andygrove (approved):

benchmarks/tpch/Cargo.toml
Dandandan: Yes, that makes sense! There are already some performance improvements (and features) after the 3.0 release, but that could be tested later as well.

andygrove (approved):
Thanks @Dandandan

benchmarks/tpch/Cargo.toml
houqp: We have the same problem with https://github.com/delta-io/delta-rs, where we basically have to stay closer to arrow and datafusion master due to various bug fixes and performance improvements. The latest big change we need to pull in is the tokio 1.0 upgrade, which makes it not possible for us to stay with 3.0.

What would be really nice is to have dev, beta or alpha release for arrow and datafusion crates in crate.io that keeps a closer track of master branch. Given the velocity of change for these two crates right now, i think this will help make crate consumption a lot easier for downstream projects until these two crates reach a fairly stable state. On the other hand, that means more work for arrow maintainers :P

andygrove: We had a discussion about this on the last Arrow Rust sync call - the next call is this Wednesday 9:30 PST - there is definitely a desire to increase release frequency and there are some ongoing discussions about how to acheive this.

jorgecarleitao: Is there an issue with using git hashes on `Cargo.toml` (apart from not looking "natural")? I find them quite explicit about what code is being used, and I have used it myself on the `python-datafusion` project.

Another thing I do there is to sometimes point to a commit on my own repo, e.g. when a PR has not been merged: this way I do not need to pressure myself nor the arrow team to merge a specific PR.

In high throughput projects, I think that it is valuable to have "stable forks".

houqp: The only problem with using git rev in Cargo.toml is crates.io won't allow you to publish it :(

We have been using git revs for delta-rs's python binding as well since pypi doesn't care about how cargo builds the binary, so our python binding release is much ahead of the core rust release.

andygrove: I created an issue to continue the discussion about Arrow dependency version / strategy 

https://github.com/ballista-compute/ballista/issues/433

PR #430, opened 2021-01-25
Implement serde for physical operator ProjectionExec

### LINKED ISSUE #426: Implement serde for physical operator ProjectionExec

bjcohen:
Fixes #426


andygrove (approved):
Thanks @bjcohen  this looks good

PR #437, opened 2021-01-27
Implement serde for Empty, LocalLimit and GlobalLimit

edrevo:


edrevo:
cc @morpheyesh. feel free to ask any question or give feedback about the code. if you get stuck with HashJoinExec we can work it out together if you want.

morpheyesh:
Nice. I get it. I will take a stab at HashJoinExec tomorrow. It is not as straightforward as Empty and Local/Global Limit but slightly similar to JoinExec. Thanks

andygrove (approved):
Thanks @edrevo LGTM

PR #440, opened 2021-01-28
Implement serde for SortExec

### LINKED ISSUE #425: Implement serde for physical operator SortExec

sieut:
Fix #425


rust/ballista/proto/ballista.proto
```diff
@@ -325,6 +326,12 @@ message LocalLimitExecNode {
   uint32 limit = 2;
 }
 
+message SortExecNode {
+  PhysicalPlanNode input = 1;
+  repeated LogicalExprNode expr = 2;
+  uint32 concurrency = 3;
```
andygrove, line 332: We can remove the concurrency field since this is specific to DataFusion and won't be needed in Ballista, which will have its own logic for determining this.

andygrove (approved):
LGTM! Thanks @sieut I left one comment

PR #445, opened 2021-01-29
Implement standalone mode scheduler client

### LINKED ISSUE #442: Implement "standalone" mode
I have partially implemented standalone cluster mode in https://github.com/ballista-compute/ballista/pull/441 and it needs completing.

We will need to add some new protobuf messages. The serde code should go in the `serde::scheduler` module.

We need one message that an executor can use to register its metadata (uuid, host, port) with a registrar.

We need a request to get all the executors registered with the registrar, and a response that lists the metadata (uuid, host, port) for each executor that is registered.

The registrar is just an executor started in a special registrar mode (will need an extra command-line arg for this).

edrevo:
Fixes #442

Once this is merged I will work on some more changes on the scheduler-client side that might be more opinionated, so I don't want to mix them with this change.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -204,7 +204,6 @@ dependencies = [
  "kube",
  "log",
  "prost",
- "prost-build",
  "prost-types",
  "snmalloc-rs",
  "sqlparser",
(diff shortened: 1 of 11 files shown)


andygrove (approved):
This looks great. Thanks @edrevo I am merging this now and will test it out locally over the weekend.

Issue #382, opened 2021-01-18
Scheduler design question: How to distribute tasks to executors?


andygrove:
The scheduler will break a query down into stages and tasks and then needs to send them to executors for execution.

The simplest approach is just to split the tasks into an equal number of chunks and then send them to executors on a round-robin basis by sending Flight actions to each executor and then wait for those requests to return a completion status (risk of network timeouts) or poll to check for status (inefficient).

One problem with this approach is that some executors may finish before others because the tasks can take variable lengths of time to complete.

Another option could be to write the tasks to etcd and implement a work queue design where each executor keeps pulling tasks until there are no more tasks. The scheduler can listen to etcd events to determine when the tasks are completed (when the queue is empty).

Using etcd would mean adding an etcd dependency to the Kubernetes deployment model.

Thoughts?

chadbrewbaker:
My deployment  would be an EC2 Batch head node that spawns Batch Spot and AWS Lambda executors.  No Kubernetes. Preferable to package binaries as AWS EC2/Lambda zipfiles.  Keep the Kubernetes provisioning optional.

The https://github.com/StanfordSNR/gg task graph scheduler seems performant.  I have a side project of using a SMT solver to job shop schedule it for optimal cost within a runtime budget.

edrevo:
Personally, I don't mind the etcd dependency that much (my guess is we'll probably need it for HA in the schedulers either way), but I'm not sure it is a requirement for a queue-based mechanism. Couldn't the scheduler implement an API to dispatch work in a queue-based manner instead of pre-assigning all of the chunks of work?

The way I see it, there are two strategies to dispatch tasks into executors:

1. The dispatching logic resides in the scheduler: for this route, I think it is better to have the workers and scheduler communicate via API. The assumption here is that centralizing the dispatch logic might allow you to perform more complicated allocations. Speculative execution comes to mind as something that could be easier to implement on the scheduler's side.
2. The dispatching logic resides in the executors: here, having an etcd interface between the workers and schedulers fits well, but I don't necessarily see it as a requirement. The assumption in this case is the inverse: the executor has better information to choose which task it wants to run. For example, data locality is something that might be easier to track on the worker's side.

Overall, I think that having a work queue in etcd as the interface between scheduler and executors for work allocation might be a less extensible design than the API.

chadbrewbaker:
[...]

PR #446, opened 2021-01-30
Scheduler scaffolding

edrevo:
@andygrove, here's my opinionated PR on how the scheduler could work. Feel free to push back anything you don't like. I am happy to iterate and change whatever is necessary.

In this PR I add a new binary "scheduler", which is the seed to implement a scheduler process. The way I see this problem space, there are 3 roles that are needed for Ballista: executors, scheduler and resource manager.

- The executors do the heavy lifting through datafusion.
- The resource manager is in charge of dynamically adding more executors when needed, or removing those that aren't necessary. This piece doesn't currently exist in the code base, and I am not adding it here. When running on k8s, the resource manager would increase/decrease the number of replicas in the executor deployment, for example. We could also support other resource managers like YARN or Mesos.
- The scheduler is in charge of work dispatching and maintaining the state of the cluster (by state I mean metadata, not the actual big data elements).

OK, so down to the PR. In this PR I am being opinionated wrt https://github.com/ballista-compute/ballista/issues/382: I am implementing the necessary infrastructure so that the scheduler and executors communicate via a gRPC API. The scheduler can then choose to save the cluster state (for now, only the `ExecutorMeta`) wherever it wants. Two implementations are provided in this PR: etcd (reusing the code that is already in place) and standalone (which ends up saved into a local file). This decision is the same thing that k8s does: although the k8s state is stored in etcd, each kubelet communicated via gRPC with the apiserver. I believe this is the most flexible architecture since iterating wrt the cluster state (for example, we might want to replace etcd with some kind of raft consensus with multiple scheduler instances) does not impact the executors.

Another thing I am keeping in mind is that, even though there are different roles and those roles could be deployed as independent processes, we do want to have a simple dev experience to try out ballista, which would mean that all roles should be able to run in a single process (or even embedded into the client, like the Spark Local). This could be achieved in the future through cargo features for the ballista library.

I'll add comments on my own code to explain the rationale behind some of the decisions.

### DIFF
--- a/rust/ballista/Cargo.toml
+++ b/rust/ballista/Cargo.toml
@@ -26,8 +26,11 @@ kube = "0.35"
 log = "0.4"
 prost = "0.6"
 prost-types = "0.6"
-sqlparser = "0.7"
+serde = "1"
+serde_json = "1"
+sled = "0.34"
 snmalloc-rs = {version = "0.2", features= ["cache-friendly"], optional = true}
+sqlparser = "0.7"
 structopt = "0.3"
 tempfile = "3"
 tokio = { version = "0.2", features = ["full"] }
--- a/rust/ballista/Cargo.toml
+++ b/rust/ballista/Cargo.toml
@@ -41,7 +41,7 @@ datafusion = "3.0.0"
 [dev-dependencies]
 
 [build-dependencies]
-prost-build = { version = "0.6.1" }
+tonic-build = { version = "0.3", features = ["prost", "transport"], default-features = false }
 
 [[bin]]
 name = "executor"
--- a/rust/ballista/src/bin/executor.rs
+++ b/rust/ballista/src/bin/executor.rs
@@ -18,45 +18,24 @@ use std::sync::Arc;
 
 use arrow_flight::flight_service_server::FlightServiceServer;
 use ballista::flight_service::BallistaFlightService;
-use ballista::scheduler::etcd::EtcdClient;
-use ballista::scheduler::k8s::KubernetesClient;
-use ballista::scheduler::standalone::StandaloneClient;
-use ballista::scheduler::SchedulerClient;
+use ballista::serde::protobuf::scheduler_grpc_client::SchedulerGrpcClient;
 use ballista::BALLISTA_VERSION;
 use ballista::{
     executor::{BallistaExecutor, ExecutorConfig},
     serde::scheduler::ExecutorMeta,
 };
-use clap::arg_enum;
 use log::info;
 use structopt::StructOpt;
 use tonic::transport::Server;
 use uuid::Uuid;
 
-arg_enum! {
-    #[derive(Debug)]
-    enum Mode {
-        K8s,
-        Etcd,
-        Standalone
-    }
-}
-
 /// Ballista Rust Executor
 #[derive(StructOpt, Debug)]
 #[structopt(name = "basic")]
 struct Opt {
-    /// discovery mode
-    #[structopt(short, long, possible_values = &Mode::variants(), case_insensitive = true, default_value = "Standalone")]
-    mode: Mode,
-
     /// Namespace for the ballista cluster that this executor will join.
     #[structopt(long)]
     namespace: Option<String>,
-
-    /// etcd urls for use when discovery mode is `etcd`
-    #[structopt(long)]
-    etcd_urls: Option<String>,
 
     /// Registrar host to register with when discovery mode is `standalone`. This is optional
     /// because it is currently possible to run in standalone mode with a single executor and
@@ -67,8 +46,8 @@ struct Opt {
     /// Registrar port to register with when discovery mode is `standalone`. This is optional
     /// because it is currently possible to run in standalone mode with a single executor and
     /// no registrar.
-    #[structopt(long)]
-    registrar_port: Option<usize>,
+    #[structopt(long, default_value = "50050")]
+    registrar_port: u16,
 
     /// Local host name or IP address to bind to
     #[structopt(long)]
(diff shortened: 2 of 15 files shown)


rust/ballista/Cargo.toml
```diff
@@ -26,8 +26,11 @@ kube = "0.35"
 log = "0.4"
 prost = "0.6"
 prost-types = "0.6"
-sqlparser = "0.7"
+serde = "1"
+serde_json = "1"
+sled = "0.34"
```
edrevo, line 31: These 3 are used for standalone mode. Sled is an embedded key-value store.

rust/ballista/src/bin/executor.rs
```diff
@@ -18,46 +18,25 @@ use std::sync::Arc;
...
 use uuid::Uuid;
 
-arg_enum! {
-    #[derive(Debug)]
-    enum Mode {
-        K8s,
-        Etcd,
-        Standalone
-    }
-}
```
edrevo, line 43: Executors no longer need to know where the cluster state is stored: they just need the URL to the scheduler/registrar.

rust/ballista/src/bin/executor.rs
```diff
@@ -94,38 +73,30 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // parse command-line arguments
     let opt = Opt::from_args();
-    let namespace = opt.namespace.unwrap_or_else(|| "ballista".to_owned());
+    let _namespace = opt.namespace.unwrap_or_else(|| "ballista".to_owned());
```
edrevo, line 76: If we add the namespace to the ExecutorMeta, then a single scheduler would be able to manage multiple clusters, which might be interesting. I haven't made that change in this PR.

rust/ballista/src/scheduler/k8s.rs
```diff
@@ -14,6 +14,7 @@
 
 //! Ballista k8s cluster management utilities
 
+/*
```
edrevo, line 17: I am not sure this implementation is a good idea, which is why I have commented it out: the fact that a pod exists isn't enough to add it to the available executors, imho. The pod could be Pending, in CrashLoopBackOff or maybe its liveness probe is failing.

On the other hand, if we want to store the cluster state in K8s, then the `ConfigBackendClient` could be built by using a ConfigMap as its backend, I guess. I'm not sure of the limitations or if that would be useful. If it is, I can create it.

rust/ballista/src/scheduler/mod.rs
```diff
@@ -18,14 +18,89 @@ pub mod etcd;
...
+};
 use crate::serde::scheduler::ExecutorMeta;
 
-use async_trait::async_trait;
+#[tonic::async_trait]
+pub trait ConfigBackendClient: Clone {
+    async fn get(&mut self, key: &str) -> Result<Vec<u8>>;
+    async fn get_from_prefix(&mut self, prefix: &str) -> Result<Vec<Vec<u8>>>;
+    async fn put(&mut self, key: String, value: Vec<u8>) -> Result<()>;
+}
```
edrevo, line 36: ConfigBackendClient is how the scheduler saves and restores state. Right now it's just executor metadata, but it could also have a list of jobs and their state and basically any kind of useful metadata about the cluster.

rust/ballista/src/bin/executor.rs
```diff
@@ -94,38 +73,30 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            let registrar_host = opt.registrar_host.unwrap_or_else(|| "localhost".to_owned());
-            let registrar_port = opt.registrar_port.unwrap_or(50051);
-            Arc::new(
-                StandaloneClient::try_new(registrar_host, registrar_port, executor_meta).await?,
-            )
-        }
-        Mode::K8s => Arc::new(KubernetesClient::new(namespace.to_owned(), namespace)),
-    };
-
+    let scheduler =
```
andygrove, line 96: It would be nice to retain the ability to start a single executor without requiring a scheduler process, especially for testing and local development. Maybe this is a new mode we need to add?

rust/ballista/src/bin/scheduler.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// etcd urls for use when discovery mode is `etcd`
+    #[structopt(long, default_value = "localhost:2379")]
+    etcd_urls: String,
+
+    /// Local host name or IP address to bind to
+    #[structopt(long, default_value = "0.0.0.0")]
+    bind_host: String,
+
+    /// bind port
+    #[structopt(short, long, default_value = "50051")]
```
andygrove, line 41: The executor tries to connect to the scheduler at 50050 by default, so we should change this value to match?

andygrove:
This looks really great @edrevo. Thank you for taking the lead on this. I tried this out locally and left a couple of comments.

Could you update the `benchmarks/tpc-h/README.md` to explain how to start the scheduler? I figured it out but it won't be obvious to a new user.

rust/ballista/src/bin/scheduler.rs
```diff
@@ -0,0 +1,80 @@
...
+    port: u16,
+}
+
+async fn start_server<T: ConfigBackendClient + Send + Sync + 'static>(
+    config_backend: T,
+    namespace: String,
+    addr: SocketAddr,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let server = SchedulerGrpcServer::new(SchedulerServer::new(config_backend, namespace));
```
andygrove, line 50: Could you add some `INFO` logging here so we can see the scheduler start up and see what port it is binding to, similar to what we have in the executor process.

rust/ballista/src/bin/executor.rs
```diff
@@ -67,8 +46,8 @@ struct Opt {
     /// Registrar port to register with when discovery mode is `standalone`. This is optional
     /// because it is currently possible to run in standalone mode with a single executor and
     /// no registrar.
-    #[structopt(long)]
-    registrar_port: Option<usize>,
+    #[structopt(long, default_value = "50050")]
+    registrar_port: u16,
```
andygrove, line 50: `registrar` should probably be renamed to `scheduler` now? Do the comments need updating as well?

andygrove (approved):
I am going ahead with merging this and we can fix the small issues with follow up PRs

rust/ballista/src/bin/scheduler.rs
edrevo: 🤦🏻‍♂️

andygrove:
@edrevo BTW, I did follow up with a couple minor fixes in separate PRs that have been merged

PR #450, opened 2021-01-30
Implement serde for physical plan Csv/Parquet scan

andygrove:


rust/ballista/proto/ballista.proto
```diff
@@ -274,13 +275,19 @@ message PhysicalPlanNode {
...
 
-message ScanExecNode {
+message ParquetScanExecNode {
+  repeated string filename = 1;
+  repeated uint32 projection = 2;
+}
+
+message CsvScanExecNode {
   string path = 1;
   repeated uint32 projection = 2;
```
edrevo, line 285: Does projection make sence for Csv?

andygrove: Yes, this is the list of columns that we want to load from the CSV file. We only want to load columns that the query needs.

PR #451, opened 2021-01-30
Upgrade Arrow

andygrove:
I ran into limitations with DataFusion 3.0.0 so we are going to need to start using the latest Arrow again.


andygrove:
@Dandandan fyi I ended up having to use latest from Arrow

PR #453, opened 2021-01-30
HashJoinExec Serde - Physical Plan

### LINKED ISSUE #429: Implement serde for physical operator HashJoinExec

morpheyesh:
Fixes #429 

Big thanks to @edrevo

### DIFF
--- a/rust/ballista/src/serde/physical_plan/from_proto.rs
+++ b/rust/ballista/src/serde/physical_plan/from_proto.rs
@@ -16,14 +16,17 @@
 
 use std::sync::Arc;
 use std::{convert::TryInto, unimplemented};
+use datafusion::physical_plan::hash_utils::JoinType;
 
 use datafusion::physical_plan::{
     empty::EmptyExec,
     limit::{GlobalLimitExec, LocalLimitExec},
     projection::ProjectionExec,
+    hash_join::HashJoinExec,
+
 };
 use datafusion::physical_plan::{ExecutionPlan, PhysicalExpr};
-
+use datafusion::physical_plan::hash_utils::JoinOn;
 use crate::error::BallistaError;
 use crate::serde::{proto_error, protobuf};
 use crate::{convert_box_required, convert_required};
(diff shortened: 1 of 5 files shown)


rust/ballista/src/serde/physical_plan/from_proto.rs
```diff
@@ -75,6 +76,28 @@ impl TryInto<Arc<dyn ExecutionPlan>> for &protobuf::PhysicalPlanNode {
                 Ok(Arc::new(LocalLimitExec::new(input, limit.limit as usize)))
             }
             PhysicalPlanType::HashAggregate(_) => unimplemented!(),
+            PhysicalPlanType::HashJoin(hashjoin) => {
+                let _left: Arc<dyn ExecutionPlan> = convert_box_required!(hashjoin.left)?;
+                let _right: Arc<dyn ExecutionPlan> = convert_box_required!(hashjoin.left)?;
```
andygrove, line 81: This should be referencing `hashjoin.right` rather than `hashjoin.left`

rust/ballista/src/serde/physical_plan/from_proto.rs
```diff
@@ -75,6 +76,28 @@ impl TryInto<Arc<dyn ExecutionPlan>> for &protobuf::PhysicalPlanNode {
                 Ok(Arc::new(LocalLimitExec::new(input, limit.limit as usize)))
             }
             PhysicalPlanType::HashAggregate(_) => unimplemented!(),
+            PhysicalPlanType::HashJoin(hashjoin) => {
+                let _left: Arc<dyn ExecutionPlan> = convert_box_required!(hashjoin.left)?;
```
andygrove, line 80: Can you remove the leading underscores from these variable names?

rust/ballista/src/serde/physical_plan/to_proto.rs
```diff
@@ -136,6 +112,31 @@ impl TryInto<protobuf::PhysicalPlanNode> for Arc<dyn ExecutionPlan> {
                     },
                 ))),
             })
+
+        } else if let Some(exec) = plan.downcast_ref::<HashJoinExec>() {
+            let _left: protobuf::PhysicalPlanNode = exec.left().to_owned().try_into()?;
```
andygrove, line 117: Please remove leading underscores from variable names

rust/ballista/src/serde/physical_plan/to_proto.rs
```diff
@@ -136,6 +112,31 @@ impl TryInto<protobuf::PhysicalPlanNode> for Arc<dyn ExecutionPlan> {
...
+            let _right: protobuf::PhysicalPlanNode = exec.right().to_owned().try_into()?;
+            let _on: Vec<protobuf::JoinOn> = exec.on()
+                .iter()
+                .map(|tple| protobuf::JoinOn {
```
andygrove, line 121: nit: can we name the variable `tuple`?

andygrove (approved):
This is looking good. Thanks @morpheyesh I left some feedback.

rust/ballista/src/serde/physical_plan/from_proto.rs
morpheyesh: Oops! Fixed it. Thanks :)

rust/ballista/src/serde/physical_plan/from_proto.rs
morpheyesh: Yes. Apologies if it's not the convention, I think I borrowed it from another variable that has the leading underscores. I will need to read up on rust. It is fixed :)

rust/ballista/src/serde/physical_plan/to_proto.rs
morpheyesh: Fixed :)

rust/ballista/src/serde/physical_plan/from_proto.rs
andygrove: So, Rust complains if you have variables that are not used anywhere, and prefixing them with `_` gets rid of that warning. Hopefully, we will get to a point soon where we no longer have any of these in the codebase.

morpheyesh: Ah, Got it 👍

PR #456, opened 2021-01-31
Fix race conditions in executor

### LINKED ISSUE #455: Race conditions between scheduler and executor startup
See https://github.com/ballista-compute/ballista/pull/452#discussion_r567330286 for context.

There are 2 race conditions between the scheduler and executor startup:

1. If the executor is running in local mode, there is no guarantee that the in-proc scheduer has started before we try to register the executor.
2. The executor will register itself with the scheduler before being ready to accept incoming requests.

edrevo:
Fixes #455

Also, it improves the docker-compose file for the tpch benchmark, but it doesn't work yet. I get the following error message: 

```
Error: General("Status { code: Internal, message: \"Ballista Error: DataFusionError(IoError(Os { code: 2, kind: NotFound, message: \\\"No such file or directory\\\" }))\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Sun, 31 Jan 2021 21:46:29 GMT\"} } }")
```

I guess it is a problem with the volume mounts, but I can't seem to get the correct path combination


edrevo:
merge conflicts fixed

rust/ballista/src/scheduler/etcd.rs
```diff
@@ -59,23 +59,13 @@ impl ConfigBackendClient for EtcdClient {
     }
 
     async fn put(&mut self, key: String, value: Vec<u8>) -> Result<()> {
-        let lease_time_seconds = 60;
-        match self.etcd.lease_grant(lease_time_seconds, None).await {
```
andygrove, line 63: If we remove the lease grant approach, we lose the ability to detect when schedulers lose contact with etcd?

edrevo: Yeah, I was tracking that through https://github.com/ballista-compute/ballista/issues/457 but I should just fix it in this PR.

edrevo:
This PR now fixes #457 too

andygrove (approved):
LGTM. Thanks @edrevo

Issue #448, opened 2021-01-30
Implement physical plan serde for these expressions


andygrove:
See `ballista/src/serde/physical_plan/to_proto.rs`

```rust
        } else if let Some(_expr) = expr.downcast_ref::<Literal>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<CaseExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<CastExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<NotExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<IsNullExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<IsNotNullExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<InListExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<InListExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<NegativeExpr>() {
            unimplemented!()
        } else if let Some(_expr) = expr.downcast_ref::<ScalarFunctionExpr>() {
            unimplemented!()
```

boaz-codota:
I'd love to contribute this issue, but I need further guidance. Like basically how can I generate the protobuf's rust representation.

boaz-codota:
I've got stuck on `ScalarFunctionExpr` & `Literal` (I have no idea what to do).

And I thought I know what to do with `CastExpr` but got some weird compilation errors (will submit a PR for help once the last PR gets merged)

andygrove:
@boaz-codota I will put up a PR for `ScalarFunctionExpr` shortly

j4ckcyw:
@boaz-codota  @andygrove  Let me take a stab at `Literal`. ;)

j4ckcyw:
Looks like Literal is already done. Doesn't hurt to add a test case for this ;)

j4ckcyw:
Actually, looks like `Literal` is already done. 

https://github.com/ballista-compute/ballista/blob/df787c9e1ecc0e5236d6e5e416e0310b6bdceff7/rust/ballista/src/serde/logical_plan/from_proto.rs#L789
[...]

PR #466, opened 2021-02-03
Implement serde for logical operators NegativeExpr & IsNotNullExpr

boaz-codota:
Ref https://github.com/ballista-compute/ballista/issues/448


andygrove (approved):
Thanks @boaz-codota

PR #467, opened 2021-02-04
Scheduler/executor refactor

andygrove:
This PR simply moves some files around and there are no code changes:

- There is now an `executor` module containing the flight service and executor code
- The distributed query planner has been moved into the `scheduler` module
- The distributed query execution code has been moved into the `executor` module


andygrove:
@edrevo fyi

edrevo (approved):

PR #479, opened 2021-02-06
Implement serde for logical operator CaseExpr

boaz-codota:
Ref #448


andygrove (approved):
LGTM. Thanks @boaz-codota

Issue #477, opened 2021-02-06
Implement ability to load configuration properties from file


andygrove:
When launching the scheduler and executor processes, we need the ability to specify a configuration file containing configuration options.  For example, `cargo run --bin executor --conf-file myconf.properties`.

The configuration file will consist of name-value pairs. For example:

```
ballista.default_batch_size=32768
ballista.feature.distributed.enabled=true
```

Shamazo:
I could take a look at this tomorrow

andygrove:
Fixed in https://github.com/ballista-compute/ballista/issues/490

PR #490, opened 2021-02-07
conf file support

Shamazo:
addresses #477 

This is implemented in a minimalist manner. Config files are expected to be key-value pairs. Values are separated from keys by '='. Each key value pair must be on a new line.

This also changes flags to use key value pairs, so we can set `local=false` in our config, and override with `--local true` on the commandline.

Options can be set with both config files and command line arguments and command line arguments will override options
set in config files. This is achieved by configuring clap/structopt to allow all params to be overridable and parsing a config file into a vector of strings in the same format as command line arguments and prepending that vector to the command line arguments. For example a conf file with
```
scheduler-port=1234
local=true
```

is parsed as `["--scheduler-port", "1234", "--local" "true"]`

which is then parsed by structopt. (With the minor detail that we prepend with the binary name, as structopt expects that)

I am fairly new to rust, so would appreciate critical feedback on how idiomatic my code is as well as my approach on this. Happy to make any changes as necessary. I also went with a KISS approach as something like the [config crate](https://docs.rs/config/0.10.1/config/) seemed possibly overkill, and harder to work with command line arguments. 

I may also need to go through anywhere the scheduler or executor are invoked to confirm that the `--local` to `--local [true|false]` change didn't break anything.


boazberman:
Hey @Shamazo, maybe we can use some pre-built crate to parse the config file?
Something like `config` crate can do the heavy lifting: https://crates.io/crates/config

rust/ballista/examples/example_executor.conf
```diff
@@ -0,0 +1,5 @@
+#this could be overridden with "--scheduler-port=5678" as a command line argument
```
andygrove, line 1: Will this work with configs where the name contains periods e.g. "ballista.csv.batch_size" ? The approach spark uses is to have a generic `--conf` command-line argument e.g. `--conf ballista.csv.batch_size = 32768`

Shamazo: yes, but the way this is implemented the config names are are exactly the same as the command line arguments parsed by structopt. So fields of the opts would have to be defined like this for example:
```
/// Local host name or IP address to bind to
    #[structopt(long="ballista.csv.batch_siz", default_value = 4096)]
    batch_size: uint32,
```

The more I look at this the less I like it. I found the [configure_me](https://crates.io/crates/configure_me) crate and I think I can write up a better implementation using it, replacing ballista's use of structopt. Structopt is great, but it is basically a wrapper around clap and neither really support config files. 

Are TOML files fine for config?

andygrove: I hadn't thought about using TOML, but that would be interesting to explore.

Shamazo: TOML seems pretty neat, working on a first pass of it right now. This should allow configuration from a TOML file, environment variables and from command line arguments.

andygrove:
Closing this because it has been replaced by https://github.com/ballista-compute/ballista/pull/535

Issue #485, opened 2021-02-07
DataFrame::collect should send logical plan to scheduler, not executor


andygrove:
Once https://github.com/ballista-compute/ballista/pull/480 is merged, the next step is to have `DataFrame::collect`send the logical plan to a scheduler's `ExecuteLogicalPlan` gRPC method and receive the list of result partitions back and then create and execute a `CollectExec` to fetch the results.

andygrove:
@edrevo I may need some help with this

PR #491, opened 2021-02-07
[WIP] Use scheduler from BallistaContext

edrevo:
Here's my progress so far for #485. Early feedback is welcome.


andygrove:
Thanks @edrevo this is looking great - we will need to do similar async work + polling for the executors at some point

edrevo:
This compiles now. Tpch is still failing with a serde error I can't pin down... basically, `serde::from_slice(serde::to_vec(JobMeta))` is failing with `Error reading job metadata: Internal Ballista error: Could not deserialize state value: missing field `metadata` at line 1 column 38`. Seems to be something related to how the `Schema` object is being serialized.

rust/ballista/src/context.rs
```diff
@@ -205,42 +210,68 @@ impl BallistaDataFrame {
...
+                    job_id: job_id.clone(),
                 })
-                .await
+                .await?
+                .into_inner();
+            if partition_location.is_empty() {
+                info!("Job {} still executing...", job_id);
+                tokio::time::sleep(Duration::from_secs(5)).await;
+            } else {
+                // TODO: use streaming. Probably need to change the signature of fetch_partition to achieve that
```
andygrove, line 247: You should be able to use `CollectExec` here to stream the final results

rust/ballista/proto/ballista.proto
```diff
@@ -429,7 +429,19 @@ message ExecuteQueryParams {
...
+
+message GetJobStatusParams {
+  string job_id = 1;
+}
+
+message GetJobStatusResult {
+  // TODO: right now we only return the partition_locations once all of them are available
+  // We should probably add a bunch more info here, like a job status (queued, running, error, done...)
+  // along with progress report for each stage
+  Schema schema = 1;
```
andygrove, line 443: The schema is known by the planner before the job is submitted. There is no need to serialize it here.

rust/ballista/src/context.rs
```diff
@@ -205,42 +210,68 @@ impl BallistaDataFrame {
...
-            }
+            format!("http://{}:{}", state.scheduler_host, state.scheduler_port)
         };
 
-        info!("Connecting to Ballista executor at {}:{}", host, port);
+        info!("Connecting to Ballista scheduler at {}", scheduler_url);
+
+        let mut scheduler = SchedulerGrpcClient::connect(scheduler_url).await?;
 
         let plan = self.df.to_logical_plan();
```
andygrove, line 224:
```suggestion
let schema: Schema = plan.schema().as_ref().clone().into();
```

rust/ballista/src/context.rs
```diff
@@ -205,42 +210,68 @@ impl BallistaDataFrame {
...
+                                &Uuid::parse_str(&partition_id.job_uuid).unwrap(),
+                                partition_id.stage_id as usize,
+                                partition_id.partition_id as usize,
+                            )
+                            .await?,
+                    );
+                }
+                break Ok(Box::pin(MemoryStream::try_new(
+                    result,
+                    Arc::new((&schema.unwrap()).try_into()?),
```
andygrove, line 271:
```suggestion
Arc::new(schema),
```

rust/ballista/src/context.rs
```diff
@@ -205,42 +210,68 @@ impl BallistaDataFrame {
...
+            .execute_logical_plan(ExecuteQueryParams {
+                logical_plan: Some((&plan).try_into()?),
+            })
+            .await?
+            .into_inner()
+            .job_id;
+
+        loop {
+            let GetJobStatusResult {
+                schema,
```
andygrove, line 235:
```suggestion
```

andygrove (commented):
I tested this locally and was able to get past the issue by removing the need to serialize the schema as part of JobMeta. The schema is known in advance of submitting the job so I don't think we need to serialize it, at least not right now.

I ran into another issue where the query failed because there were no executors available, causing a divide by zero in this line in planner.rs

```rust
let executor_meta = &executors[child_partition % executors.len()];
```

rust/ballista/src/scheduler/mod.rs
```diff
@@ -127,77 +107,132 @@ impl<T: ConfigBackendClient + Send + Sync + 'static> SchedulerGrpc for Scheduler
...
+                let plan = datafusion_ctx
+                    .optimize(&plan)
+                    .and_then(|plan| datafusion_ctx.create_physical_plan(&plan))
+                    .map_err(|e| {
+                        let msg =
+                            format!("Could not retrieve data from configuration store: {}", e);
+                        error!("{}", msg);
+                        tonic::Status::internal(msg)
+                    })
+                    .unwrap();
```
andygrove, line 173: This panicked for me. We should catch the error and mark the job as failed here.

edrevo: Yeah, error management isn't baked in yet. I have a TODO comment to have stuff like "job status" which would allow marking a job as failed. I'll do that in another PR if that OK, to avoid huge PRs

edrevo:
Progress! When you now run the tpch benchmark it still fails, but it actually computed the result! Here's what's happening:

1. The context (tpch) is connecting to the scheduler and sending the LogicalPlan
2. The scheduler converts the LogicalPlan into a distributed physical plan via the planner, sends the QueryStage to the executor and returns a ShuffleReaderExec node.
3. The scheduler extracts the partition_locations from the ShuffleReaderExec and sends them back to the context.
4. The context receives the partition locations once the jobs finishes and tries to read them back through ballista_client.fetch_partition
5. The executor receives the FetchPartition action:

```
[2021-02-09T16:09:59Z INFO  ballista::executor::flight_service] FetchPartition PartitionId { job_uuid: eac64c54-d12e-4726-9f50-3e56e5b9eb25, stage_id: 2, partition_id: 0 }
[2021-02-09T16:09:59Z INFO  ballista::executor::flight_service] FetchPartition PartitionId { job_uuid: eac64c54-d12e-4726-9f50-3e56e5b9eb25, stage_id: 2, partition_id: 0 } reading /tmp/.tmpUdAn5n/eac64c54-d12e-4726-9f50-3e56e5b9eb25/2/0/data.arrow
```

and finally... the executor returns an `ArrowError: IoError("No such file or directory (os error 2)")`. That was close!

Not sure why the executor isn't able to read the partition. Maybe the job_uuid isn't stable and is being generated in different places? The `/tmp/.tmpUdAn5n` directory exists and it is empty, so maybe the executor just isn't saving anything to disk.

edrevo:
I just looked up the code for QueryStageExec. Isn't it supposed to save the results to disk?

edrevo:
Ah! Found the problem:

```
        // TODO: this won't compile because it causes the resulting future to be !Sync
        /*
        let mut client = BallistaClient::try_new(&executor_meta.host, executor_meta.port as usize)
            .await
            .map_err(|e| DataFusionError::Execution(format!("Ballista Error: {:?}", e)))?;

        let _partition_metadata = client
            .execute_partition(*job_uuid, stage_id, child_partition, plan.clone())
            .await
            .map_err(|e| DataFusionError::Execution(format!("Ballista Error: {:?}", e)))?;
            */
```

edrevo:
I'll take a look at this tonight. The !Sync problem comes from FlightServiceClient and looks related to https://github.com/hyperium/tonic/issues/241

Using a channel is not ideal, but it might be the only option for now.

edrevo:
We're almost there! We're just missing the necessary serdes:

```
[2021-02-09T17:11:34Z ERROR ballista::scheduler] Could not execute distributed plan: DataFusion error: Execution("Ballista Error: General(\"physical plan to_proto unsupported plan HashAggregateExec
```

rust/ballista/src/scheduler/mod.rs
```diff
@@ -127,77 +107,132 @@ impl<T: ConfigBackendClient + Send + Sync + 'static> SchedulerGrpc for Scheduler
...
             })?;
 
-            // create physical plan using DataFusion
-            let datafusion_ctx = ExecutionContext::new();
-            let plan = datafusion_ctx
-                .optimize(&plan)
-                .and_then(|plan| datafusion_ctx.create_physical_plan(&plan))
+            debug!("Received plan for execution: {:?}", plan);
+
+            let job_id: String = {
```
andygrove, line 134: Any reason not to use UUID here?

andygrove (approved):
LGTM @edrevo please feel free to merge when you are happy and we can follow up with PRs for error handling.

There is a WIP PR up for serde for HashAggregateExec so once that is merged we can see what else we run into.

rust/ballista/src/scheduler/mod.rs
edrevo: Uuids are long and user-unfriendly imho. I am glad k8s doesn't use uuids for pod names, for example.

This is completely opinionated, so I can use Uuids if you want.

edrevo:
I'm going to add a bit of error handling first. If we merge this then the tpch benchmark will be broken. Are you ok with that?

edrevo:
I've also changed serialization in etcd to use protobuf, so we don't depend on serde/serde_json anymore. Feel free to merge if everything looks good to you.

The errors in the execution now get propagated to the client/context, so running the tpch will give you this:

```log
[2021-02-10T16:57:40Z INFO  ballista::context] Connecting to Ballista scheduler at http://localhost:50050
[2021-02-10T16:57:40Z INFO  ballista::context] Job LdRgerm is running...
[2021-02-10T16:57:45Z ERROR ballista::context] Job LdRgerm failed: status: Internal, message: "Could not execute distributed plan: DataFusion error [...]
```
```

**Comment by edrevo** [comment]
OK, I implemented a bunch of missing serde stuff and the tpch now errors with this message:

```
[2021-02-10T21:55:05Z ERROR ballista::scheduler] Could not execute distributed plan: DataFusion error: Execution("Ballista Error: General(\"Status { code: Internal, message: \\\"Ballista Error: DataFusionError(ArrowError(InvalidArgumentError(\\\\\\\"Unable to get field named \\\\\\\\\\\\\\\"l_quantity\\\\\\\\\\\\\\\". Valid fields: [\\\\\\\\\\\\\\\"l_returnflag\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"l_linestatus\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"SUM(l_quantity)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"SUM(l_extendedprice)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"SUM(l_extendedprice Multiply Int64(1) Minus l_discount)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"SUM(l_extendedprice Multiply Int64(1) Minus l_discount Multiply Int64(1) Plus l_tax)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"AVG(l_quantity)[count]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"AVG(l_quantity)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"AVG(l_extendedprice)[count]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"AVG(l_extendedprice)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"AVG(l_discount)[count]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"AVG(l_discount)[sum]\\\\\\\\\\\\\\\", \\\\\\\\\\\\\\\"COUNT(UInt8(1))[count]\\\\\\\\\\\\\\\"]\\\\\\\")))\\\", metadata: MetadataMap { headers: {\\\"content-type\\\": \\\"application/grpc\\\", \\\"date\\\": \\\"Wed, 10 Feb 2021 21:55:05 GMT\\\"} } }\")")
```

It looks like this is the reply of the executor to a do_get request right after computing the result of the plan:

```
[2021-02-10T21:55:05Z INFO  ballista::executor::flight_service] Executed partition in 111 seconds. Statistics: PartitionStats { num_rows: 4, num_batches: 1, num_bytes: 2032, null_count: 0 }
[2021-02-10T21:55:05Z INFO  ballista::executor::flight_service] Received do_get request
```

Looks like the correct column name is `SUM(l_quantity)`, but I'm not sure why the scheduler is asking for `l_quantity`. Maybe it is a problem when generating the `ShuffleReaderExec`? I'll leave it here for today.

**Comment by andygrove** [comment]
> Looks like the correct column name is `SUM(l_quantity)`, but I'm not sure why the scheduler is asking for `l_quantity`. Maybe it is a problem when generating the `ShuffleReaderExec`? I'll leave it here for today.

I'd guess that this is related to the two-step process for aggregate queries where we do a Partial aggregate first and then do a Final aggregate on those results. I'll try and find time to look tonight. It would be worth trying a query that doesn't have an aggregate to see if that works.

**Comment by andygrove** [comment]
I took a quick look at this tonight and confirmed that the partial hash aggregates appear to execute correctly and then the final hash aggregate fails. I don't know if this is a trivial issue due to incorrect schema or whether it will require changes in DataFusion and I won't have time to dig deeper until the weekend, unfortunately.

I did confirm that non-aggregate queries work just fine.

**Comment by edrevo** [comment]
Thanks for the comment. I'll try to dig some more today

File: rust/ballista/src/scheduler/mod.rs (hunk 7 above)
Comment (andygrove) [reply rust/ballista/src/scheduler/mod.rs:134]: Sure, that makes sense.

**Comment by edrevo** [comment]
I don't have more time today, but hopefully this is useful information. Stage 0 gets executed correctly, and Stage 1 fails when trying to deserialize the protobuf into an ExecutionPlan. I think the problem comes from this snippet:

```
SortExec {
    input: ProjectionExec {
        expr: [ ... ],
        schema: Schema { ... },
        input: HashAggregateExec {
            mode: Final,
            group_expr: [ ... ],
            aggr_expr: [
                Sum { name: \"SUM(l_quantity)\", data_type: Float64, expr: Column { name: \"l_quantity\" }, nullable: true },
                ...
            ],
            input: ShuffleReaderExec {
                partition_location:  [ ... ],
                schema: Schema {
                    fields: [
                        Field { name: \"SUM(l_quantity)[sum]\", data_type: Float64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: None },
                        ...
                    metadata: {} }
            },
            schema: Schema {
                fields: [
                    Field { name: \"SUM(l_quantity)\", data_type: Float64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: None },
                    ...
                metadata: {} }
        }
    },
    expr: [ ... ],
    concurrency: 16
}
```

If I'm understanding the execution plan correctly, the HashAggregateExec wants to produce a column named `SUM(l_quantity)` by using an input column called `l_quantity`, but if you look at its input's schema the column is really called `SUM(l_quantity)[sum]`, so that creates a mismatch which fails to parse when creating the ExecutionPlan.

The thing is, in order to create that protobuf an ExecutionPlan with that information had to exist, so I'm inclined to think this is at least a partial bug in Datafusion, since it allowed the creation of that ExecutionPlan on the scheduler's side and it won't let us re-create it on the executor's side.

**Comment by edrevo** [comment]
I just confirmed that the original datasource execution plan also has this naming mismatch, so it was not introduced by ballista

**Comment by andygrove** [comment]
So DataFusion presents a different schema for the partial aggregate:

```rust
/// returns the name of the state
pub fn format_state_name(name: &str, state_name: &str) -> String {
    format!("{}[{}]", name, state_name)
}
```

I'm going to take a look in the next couple hours and see if I at least find a workaround.

andygrove:
The issue is that the partial aggregate can output more than one accumulator per field. For example, for AVG:

```rust
    fn state_fields(&self) -> Result<Vec<Field>> {
        Ok(vec![
            Field::new(
                &format_state_name(&self.name, "count"),
                DataType::UInt64,
                true,
            ),
            Field::new(
                &format_state_name(&self.name, "sum"),
                DataType::Float64,
                true,
            ),
        ])
    }
```

andygrove:
Filed an issue/question with Arrow: https://issues.apache.org/jira/browse/ARROW-11606

jorgecarleitao:
Hi,

TL;DR: I still do not understand why we are requesting `l_quantity` on final, but here is some info.

The general schema flow for a query with an aggregation `AGG` is

* partial: `c -> (AGG(c)[state1], AGG(c)[state2], ...)`
* final: `(AGG(c)[state1], AGG(c)[state2], ...) -> AGG(c)`

e.g.

* partial: `c` -> (`AVG(c)[sum]`, `AVG(c)[count]`)
* final: (`AVG(c)[sum]`, `AVG(c)[count]`) -> `AVG(c)`

Irrespectively of the number of states that the accumulator needs to store, there are 5 types of expressions that we must physically compute:

1. physical expression to perform aggregations on (`col("c")`)
2. partial aggregations over incoming expression (`sum("c")` and `count("c")`)
3. expression to read states (`col("AVG(c)[sum]")` and `col("AVG(c)[count]")`)
4. aggregation over outgoing physical expression (`a = sum("AVG(c)[sum]")` and `b = sum("AVG(c)[count]")`)
5. final expression (`a / b`)

Step 1-2 are done in `partial` and the results are shipped over the network. Step 3-5 are done in `final` and shipped over the network.

`HashAggregateExec` currently stores `aggr_expr: Vec<Arc<dyn AggregateExpr>>` because that is sufficient to build the final expressions: in DataFusion, the columns to be read on `final` (step 3.) are constructed during execution, see `hash_aggregate::aggregate_expressions` and `hash_aggregate::merge_expressions`.

From the plan presented in debug above, `aggr_expr` points to `\"SUM(l_quantity)\"`. This is expected: during execution, the actual expressions to read (1. or 3. above) will be created from `aggr_expr` and `mode`. So, I do not see an issue there. 

We could eagerly compute the expressions to be evaluated on `HashAggregateExec::try_from` for extra information in debug. See https://github.com/jorgecarleitao/arrow/pull/25/files with a small change to DataFusion that enables that.

edrevo:
merging to unblock progress. we can track the hash aggregate issue through https://github.com/ballista-compute/ballista/issues/504.

thanks everyone for the help!

PR #483, opened 2021-02-07
WIP: HashAggregateExec

morpheyesh:
This is WIP PR. 
There isn't much added really. 
I am not entirely sure how to handle `AggregateExpr` in `to_proto`. 
Thanks for the help!

(Issue Link: #427 & #461 )

PR #497, opened 2021-02-10
Finish implementing HashAggregateExec serde

andygrove:
This extends https://github.com/ballista-compute/ballista/pull/483 and gets it to a point where the test passes. This required some knowledge of DataFusion internals to complete.

Thanks for the work on this @morpheyesh


morpheyesh:
Thanks @andygrove 👍

Issue #418, opened 2021-01-24
Implement Python bindings


andygrove:
DataFusion has Python bindings available [1] so it should be possible to adapt this code and add Python bindings for Ballista.

[1] https://pypi.org/project/datafusion/

djKooks:
@andygrove interested in here also. How do you think to start in?

andygrove:
@jorgecarleitao if you have time, would you be able to offer guidance on where to get started with this task?

jorgecarleitao:
Hey @djKooks ! Great to hear!

So, if I understand correctly, we would like something like

```python
ctx = ballista.local_context()

df = ctx.read_parquet(path: str)

df.select(...).agg(['a'], count('b'))

batches = df.collect()
```

as an initial goal.

My experience with `python-datafusion` and `pyo3` is that we will need to create structs that essentially bridge the Rust client and Python. For example, for datafusion's `ExecutionContext`, I did it like [this](https://github.com/jorgecarleitao/datafusion-python/blob/master/src/context.rs#L22).

The second step is to offer methods that map a python method to the corresponding rust method. For example

[here](https://github.com/jorgecarleitao/datafusion-python/blob/master/src/context.rs#L47):

```rust
fn create_dataframe(
        &mut self,
        partitions: Vec<Vec<PyObject>>,
        py: Python,
    ) -> PyResult<dataframe::DataFrame> {...}
```

The important part to notice here is that both the arguments and the output are python-like objects (that `pyo3` maps to rust and vice-versa).

For the DataFrame API itself, I had to create a small DataFrame API in Python that maps to the rust DataFrame API, [here](https://github.com/jorgecarleitao/datafusion-python/blob/master/src/expression.rs#L10).

My suggestion is to fork python-datafusion, that already has CI, packaging, tests, and all that stuff, modify `Cargo.toml` to depend on `Ballista`, and replace the execution Context from DataFusion by the Ballista one, fixing compilation errors until it runs. It may just work ^_^

If this works, I can try to split python-datafusion in two parts or try to publish it as a Rust lib that can be consumed by ballista. One idea is to introduce a `ExecutionContext` trait that can be plugged in on generics, so that we do not need to copy-paste code around.

djKooks:
@djKooks @andygrove thanks a lot~ I'll leave comment in slack channel if there's any question 🙏

andygrove:
Fixed in https://github.com/ballista-compute/ballista/pull/500

PR #500, opened 2021-02-11
Initial implementation of python bindings

pjmore:
Addresses #418

This is an initial implementation for ballista python bindings. It contains most of the methods implemented on the ballista dataframe, notably the join method is missing. I also added some CI testing for the python bindings and added path filters to both the existing rust CI and the added python CI. I also changed all the windows CI to cache the openssl build.

How exhaustive should this API be? Do we want to expose all of the same logical plan API that exists in rust, such as expressions and scalar literals, or does a smaller version make more sense?


python/README.md
```diff
@@ -4,4 +4,23 @@ This is a placeholder for Python bindings for Ballista.
 
 Python bindings already exist for DataFusion [1] so it should be possible to adapt this code to work with Ballista.
 
-[1] https://pypi.org/project/datafusion/
\ No newline at end of file
+[1] https://pypi.org/project/datafusion/
+
+Currently the easiest way to expirament with the bindings is to activate a virtual environment and then run maturin develop
+
+This is currently not manylinux compliant due to ballista's k8s dependency which uses hyper-tls which relies on libcrypto.so.1.1 through openssl. Options are to use features on the main ballista library and make it an optional dependency for building executor/scheduler or to ask for feature to vendor openssl or use rust-tls feature of k8s crate.
```
andygrove, line 11: We should make k8s support an optional feature. I will file an issue for that.

andygrove (approved):
This is awesome. Thank you @pjmore ! :tada:

PR #506, opened 2021-02-12
Slim down dependencies

### LINKED ISSUE #501: Make k8s support an optional feature in the ballista crate
It would be good to make k8s support an optional feature in the ballista crate so that we don't have to bring in all those dependencies when we are not targeting k8s for deployment.

edrevo:
Remove unnecessary dependencies in the projects to improve compile times. Fixes #501


edrevo:
Build now consists of 310 crates instead of 380

andygrove (approved):

andygrove:
@edrevo This one needs a rebase

edrevo:
looks like python bindings are already broken in main. I've opened https://github.com/ballista-compute/ballista/issues/525 to track this

PR #534, opened 2021-02-14
Refactor query stage generation

edrevo:
This PR refactors the query stage generation code. Why? Right now, the stage generation and execution happen all at once and in-memory for one of the schedulers. My goal here is moving us closer to a place where jobs can outlive both executor and scheduler failures. This PR doesn't do that, but it sets the necessary groundwork for future PRs in that direction.

The query stage generation process now returns a vector of the different stages. My future goal is to decouple stage generation from stage execution. The returned stages could be saved in etcd and let the work dispatching process happen asynchronously (potentially in another scheduler) [\*].

Having each stage as an independent unit also allows us a first take on job progress: each stage could be marked as completed and the scheduler could report that to the client [\*]. It also makes it easier to implement concurrent stage execution for those stages that can run in parallel [\*].

Query 12 runs correctly after these changes, with the same performance as before.

[\*]: not implemented in this PR


rust/ballista/src/scheduler/execution_plans/unresolved_shuffle.rs
```diff
@@ -0,0 +1,79 @@
...
+    physical_plan::RecordBatchStream,
+};
+use log::info;
+
+/// UnresolvedShuffleExec represents a dependency on the results of several QueryStageExec nodes which haven't been computed yet.
+///
+/// An ExecutionPlan that contains an UnresovledShuffleExec isn't ready for execution. The presence of this ExecutionPlan
+/// is used as a signal so the scheduler knows it can't start computation on a specific QueryStageExec.
+#[derive(Debug, Clone)]
+pub struct UnresovledShuffleExec {
```
andygrove, line 22: typo: `UnresovledShuffleExec` -> `UnresolvedShuffleExec`

rust/ballista/src/scheduler/planner.rs
```diff
@@ -94,75 +94,110 @@ impl DistributedPlanner {
...
     }
 
-    /// Insert [QueryStageExec] nodes into the plan wherever partitioning changes
-    pub fn prepare_query_stages(
+    /// Returns a vector of ExecutionPlans, where the root node is a [QueryStageExec].
+    /// Plans that depend on the input of other plans will have leaf nodes of type [UnresovledShuffleExec].
+    /// A [QueryStageExec] is created whenever the partitioning changes.
+    ///
+    /// Returns an empty vector if the execution_plan doesn't need to be sliced into several stages.
+    pub fn calculate_query_stages(
```
andygrove, line 111: Perhaps `plan_query_stages` would be a good name for this function?

andygrove (commented):
This looks good. I am going to pull this branch and test it out today to make sure that I understand how it works.

andygrove:
Sorry @edrevo but this needs rebasing now. Once this is rebased I will test locally. I'm excited for concurrent stage execution and job progress tracking.

edrevo:
Merge conflicts fixed

andygrove (approved):
Thanks @edrevo I ran the integration tests after merging this to my local branch and it looks good.

PR #535, opened 2021-02-14
implement configuration using toml, environment variables and command line

Shamazo:
we are making use of the configure_me crate. This allows for parsing of toml files, environment variables and command line arguments. In increasing order of precedence: default config file < environment variables < specified config file < command line arguments.

Todo:

- [x] write documentation
- [ ] ~remove serde dependency / implement from string for `ConfigBackend` (also removes clap dependency)~ , the first is necessary and the second saves a substantial amount of code. 
- [x] figure out how to make clippy ignore generated code properly (will probably have to write my own macro for including generated code instead of using the provided one I think)
- [x] fix docker build


andygrove:
Thanks @Shamazo I am at the end of my day but will review this tomorrow

Shamazo:
Sounds good, I'll have some time to work on this tomorrow.

Shamazo:
I think I fixed docker, I had changed build.rs and it was looking for the toml spec files, which don't exist during dependency building, so I made dummy ones and then force build.rs to rerun when we compile ballista itself.  `cargo:rerun-if-changed=` was acting wonky in the docker build environment, so I added an env var `FORCE_REBUILD`, to force it to rerun.

rust/ballista/src/bin/executor_config_spec.toml
```diff
@@ -0,0 +1,62 @@
+[general]
```
andygrove, line 1: I really like this approach. It will make it easier to generate documentation for the user guide for all the settings.

andygrove:
Thanks @Shamazo I think this is looking really good. It looks like there are some clippy issues to be resolved and you may need to rebase / merge latest from main to pick up some recent changes but LGTM

Shamazo:
example `--help` output:
```
     Running `target/debug/executor --help`
Usage: target/debug/executor [ARGUMENTS...]

Arguments:
        --config-file             Load configuration from this file.
        -n, --namespace           Namespace for the ballista cluster that this 
                                  executor will join. yippee
        --scheduler-host          Scheduler host
        --scheduler-port          scheduler port
        --bind-host               Local IP address to bind to.
        --external-host           Host name or IP address to register with 
                                  scheduler so that other executors can connect 
                                  to this executor.
        -p, --port                bind port
        --work-dir                Directory for temporary IPC files
        -c, --concurrent-tasks    Max concurrent tasks.
        --version                 Print version of this executable
        --local                   Running in local mode will launch a standalone
                                  scheduler inside the executor process. This 
                                  will create a single-executor cluster, and is 
                                  useful for development scenarios.

```

andygrove (approved):
LGTM. Thanks @Shamazo

Issue #519, opened 2021-02-13
Implement physical to_proto for Sort expression


andygrove:

hmakait:
I'll take a stab :)

PR #544, opened 2021-02-14
Serde for physical sort expr

hmakait:
This PR adds a test that shows serde for `PhysicalSortExpr` being handled within `SortExec`. Unless there is another place where `PhysicalSortExpr` is used, I think #519 can be closed.


andygrove (approved):
Thank you. I appreciate the test being added. I have been tested queries with SORT and they do work, so I agree we can close the issue.

Issue #518, opened 2021-02-13
Implement logical to_proto for ScalarFunction


andygrove:
The code largely already exists in physical to_proto and needs refactoring to move some of it to logical to_proto and have the physical to_proto use try_into

hmakait:
Working on a PR

Issue #473, opened 2021-02-06
BallistaContext should get file metadata from scheduler


andygrove:
Currently the BallistaContext simply wraps the DataFusion context, and any reference to Parquet/CSV files results in looking on the local file system for those files:

```rust
/// Create a DataFrame representing a Parquet table scan
    pub fn read_parquet(&self, path: &str) -> Result<BallistaDataFrame> {
        // convert to absolute path because the executor likely has a different working directory
        let path = PathBuf::from(path);
        let path = fs::canonicalize(&path)?;

        // use local DataFusion context for now but later this might call the scheduler
        let mut ctx = ExecutionContext::new();
        let df = ctx.read_parquet(path.to_str().unwrap())?;
        Ok(BallistaDataFrame::from(self.state.clone(), df))
    }
```

We should instead have the context send a gRPC request to the scheduler for the file metadata. This would allow a Ballista client to run against a remote cluster e.g. in the cloud and would be a step towards eventually supporting S3 and similar object stores.

hmakait:
Interested in tackling this one! :)

hmakait:
It looks like this issue should be based on #491 to avoid duplicating the integration of the context with the scheduler. Is it safe to proceed based on the existing branch, or should I wait? @andygrove

edrevo:
@hmakait, the branch has been merged, so feel free to code away!

PR #546, opened 2021-02-14
Implement logical to_proto for ScalarFunction

hmakait:
Addresses #518.

Refactors matching logic into `TryInto<protobuf::ScalarFunction> for &BuiltinScalarFunction` and uses this in both logical and physical serialization of scalar functions.

Adds a roundtrip test for `sqrt` to check implementation.

_Note: Couldn't run integration test on Mac OSX, will have to finish #473 for that._


andygrove (approved):
LGTM. Thanks @hmakait

PR #547, opened 2021-02-14
Fix tpch in docker-compose

### LINKED ISSUE #530: Scheduler cannot connect to executor in docker-compose
Integration tests are now working, with an executor running in local mode in the tpch docker-compose yaml. When running with separate scheduler and executor, the executor was registering itself as "localhost" with the scheduler.

We need to figure out how to register the correct external IP address when running in docker compose.

edrevo:
Not the best solution, since it involves compiling tpch on every run, but it beats having a broken docker-compose.

Fixes #530

### DIFF
--- a/dev/integration-tests.sh
+++ b/dev/integration-tests.sh
@@ -4,22 +4,8 @@ set -e
 pushd rust/benchmarks/tpch
 ./tpch-gen.sh
 
-# hack to make benchmark and scheduler paths identical until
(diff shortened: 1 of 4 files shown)


andygrove (approved):
LGTM. We should update `/dev/integration-tests.sh` to reflect these changes (either in this PR or as a follow on)

edrevo:
Integration tests fixed. Merging.

Issue #510, opened 2021-02-13
FetchPartition should stream results


andygrove:
The code in `flight_service.rs` for `FetchPartition` currently loads the partition from disk into a vec and then streams the results to the client. We should be able to stream the data without loading into the intermediate vec.

PR #555, opened 2021-02-15
Implement streaming in FetchPartition

andygrove:
Closes https://github.com/ballista-compute/ballista/issues/510


andygrove:
@j4ckcyw fyi

Issue #549, opened 2021-02-15
Produce optimized binaries for releases


andygrove:
According to the benchmark documentation, adding these flags will improve performance.

```
RUSTFLAGS='-C target-cpu=native -C lto -C codegen-units=1 -C embed-bitcode'
```

We should consider having a version of the Docker image that is built with these flags and use that version when running benchmarks.

Dandandan:
@andygrove for the docker image I think you don't want to compile with `target-cpu=native` as it makes the binary non-portable. It is more something to document for people wanting to build ballista/DataFusion from source.

Adding `SIMD` + `snmalloc` features (or mimalloc) and the lto stuff do improve performance though.

Dandandan:
I might have some time for this tomorrow.

PR #558, opened 2021-02-16
Add snmalloc feature to docker image

Dandandan:
This enables the snmalloc feature, addressesing a part of https://github.com/ballista-compute/ballista/issues/549
enabling simd would require a nightly rust. 

I couldn't yet figure out how to do lto, as it is not directly building a binary with the current setup. "error: lto can only be run for executables, cdylibs and static library outputs". I think we only want to have lto enabled on the executor.

### DIFF
--- a/docker/rust.dockerfile
+++ b/docker/rust.dockerfile
@@ -26,8 +26,10 @@ COPY rust/ballista/build.rs /tmp/ballista/
 RUN true
 COPY rust/ballista/src/ /tmp/ballista/src/
 # force build.rs to run to generate configure_me code.
+RUN apt-get -y install cmake
 ENV FORCE_REBUILD='true'
-RUN cargo build $RELEASE_FLAG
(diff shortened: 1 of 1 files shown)


edrevo:
LGTM

> I couldn't yet figure out how to do lto, as it is not directly building a binary with the current setup. "error: lto can only be run for executables, cdylibs and static library outputs". I think we only want to have lto enabled on the executor.

I think at some point we should move from the current cargo structure to a real multi-module structure: scheduler, executor, lib and benchmarks. That would allow enabling fat lto for the binaries. @andygrove, what do you think?

edrevo (approved):

Dandandan:
Thanks @edrevo to me that sounds like a good idea. Would be great to support/use LTO, as it's such an "easy win".

andygrove:
Thanks for this @Dandandan 

@edrevo Yes, it would be good to have separate crates and docker images for the main components.

PR #562, opened 2021-02-17
Fix compile error in python bind module

djKooks:
There were several build error in python module in newest branch:
- version difference of `arrow`
- port number type


andygrove (approved):
Thanks @djKooks

Issue #563, opened 2021-02-17
Performance issue with parquet data sources with many partitions


andygrove:
The DataFusion method `ParquetExec::try_with_files` can take several seconds to scan all of the parquet files and build statistics. When we serialize plans and send them from the scheduler and executor, we end up calling this same method for each partition in the executor and duplicate this process, and it can end up costing several minutes.

We should instead persist the partition information and statistics and use the `ParquetExec::new` constructor when deserializing the plans.

Issue #557, opened 2021-02-15
ExecutePartitions needs to return statistics to client


andygrove:
The executor code for `ExecutePartition` has statistics about the executed partition (number of rows, batches, etc) and we need to include this data in the result set returned so that the planner can use this information to make decisions when executing future query stages (such as swapping left/right side of joins)

Shamazo:
I can have a shot at this

PR #565, opened 2021-02-18
first pass at returning partition stats to scheduler

Shamazo:
Addresses #557
So far I have stats being returned in the ExecutePartitionResult object. I am not sure how the scheduler should be maintaining state, but I would love to hear thoughts. 
Not sure of the best way to test this. Going to see how it does in CI. 

I may look into writing some integration testing infrastructure in general for Ballista, but not for this issue as it is blocking #563.

### DIFF
--- a/rust/ballista/src/utils.rs
+++ b/rust/ballista/src/utils.rs
@@ -19,6 +19,10 @@ use crate::error::{BallistaError, Result};
 use crate::memory_stream::MemoryStream;
 
 use crate::scheduler::execution_plans::{QueryStageExec, UnresolvedShuffleExec};
+use arrow::array::{
+    ArrayBuilder, ArrayRef, StructArray, StructBuilder, UInt64Array, UInt64Builder,
+};
+use arrow::datatypes::{DataType, Field};
 use arrow::ipc::reader::FileReader;
 use arrow::ipc::writer::FileWriter;
 use arrow::record_batch::RecordBatch;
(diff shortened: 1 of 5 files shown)


rust/ballista/src/utils.rs
```diff
@@ -33,14 +37,109 @@ use datafusion::physical_plan::merge::MergeExec;
...
+        field_builders.push(Box::new(num_bytes_builder) as Box<dyn ArrayBuilder>);
+
+        let mut null_count_builder = UInt64Builder::new(1);
+        null_count_builder.append_value(self.null_count).unwrap();
+        field_builders.push(Box::new(null_count_builder) as Box<dyn ArrayBuilder>);
+
+        let mut struct_builder = StructBuilder::new(vec![self.arrow_struct_repr()], field_builders);
+        Arc::new(struct_builder.finish())
+    }
+
```
Shamazo, line 106: I am just learning Arrow now, would appreciate feedback on if there is a more idiomatic way to do this, seems like if we add more stats it will be annoying to maintain.

andygrove: I agree. It is a bit unusual that we are using RecordBatch to return these statistics and we should really move the `ExecutePartition` code out of `do_get` and into `do_action` so that we can just return a serialized protobuf with the statistics instead of creating a RecordBatch, but that is a larger change so we should do that as a follow up, IMO.

andygrove:
Thanks @Shamazo I have a  good idea of how this will integrate with the planner and can write up some notes or start a PR for that this weekend.

It looks like you need to fix a clippy issue to fix CI.

For testing, you could try running the [integration tests](https://github.com/ballista-compute/ballista/blob/main/docs/integration-testing.md) locally.

Shamazo:
I got them running and found a bug I need to fix (in addition to clippy), I'll get to that I'm the next few days. 

I think I am going to look into writing some more comprehensive testing infrastructure after this merges. (In my day job I'm currently on the developer productivity team so hopefully I can leverage some of my experience)

Shamazo:
I am hitting this:
```
 InvalidArgumentError("column types must match schema types, expected
 Struct([
 Field { name: \"num_rows\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None },
 Field { name: \"num_batches\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None }, 
 Field { name: \"num_bytes\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None }, 
 Field { name: \"null_count\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None }]) 
but found 
Struct([Field { name: \"partition_stats\", 
data_type: Struct([
    Field { name: \"num_rows\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None }, 
    Field { name: \"num_batches\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None},
    Field { name: \"num_bytes\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None },
    Field { name: \"null_count\", data_type: UInt64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None }]),
nullable: false, dict_id: 0, dict_is_ordered: false, metadata: None }]) at column index 1")', ballista/src/executor/flight_service.rs:143:82 
```

Been debugging why tonight, but going to have to come back to it on the weekend. Should be obvious from the diff where, I am probably doing something stupid with Arrow.

andygrove:
I can help with this. I will try and find time tomorrow.

rust/ballista/src/utils.rs
```diff
@@ -33,14 +37,109 @@ use datafusion::physical_plan::merge::MergeExec;
...
+        let mut num_bytes_builder = UInt64Builder::new(1);
+        num_bytes_builder.append_value(self.num_bytes).unwrap();
+        field_builders.push(Box::new(num_bytes_builder) as Box<dyn ArrayBuilder>);
+
+        let mut null_count_builder = UInt64Builder::new(1);
+        null_count_builder.append_value(self.null_count).unwrap();
+        field_builders.push(Box::new(null_count_builder) as Box<dyn ArrayBuilder>);
+
+        let mut struct_builder = StructBuilder::new(vec![self.arrow_struct_repr()], field_builders);
```
andygrove, line 103: @Shamazo I think this line is the issue. The data type you are passing in here is a struct but you should just be passing in the vec of individual field builders.

Shamazo:
Happy to merge this as is. We need to figure out how to maintain state when #574 merges

andygrove (approved):
LGTM! Thanks @Shamazo

Issue #513, opened 2021-02-13
Make CollectExec concurrent and stream results without intermediate Vec


andygrove:
Make CollectExec concurrent and stream results without intermediate Vec.

See TODO comments in code

j4ckcyw:
let me see.

https://github.com/ballista-compute/ballista/blob/dda19d7c74364210d7455cf64fe2c15f6fda20d6/rust/ballista/src/executor/collect.rs#L71-L72

PR #572, opened 2021-02-19
Make CollectExec concurrent and stream results without intermediate Vec

j4ckcyw:
Use the select_all stream combinator to select across a vector of
SendableRecordBatchStream

closes https://github.com/ballista-compute/ballista/issues/513


j4ckcyw:
integration-test.sh passes

andygrove (approved):
Awesome! Thanks @j4ckcyw

PR #573, opened 2021-02-19
Remove unwrap

### LINKED ISSUE #514: Remove unwrap usage from serde module
Remove unwrap usage from serde module where possible, and document why it is safe in cases where we can't remove

zynaxsoft:
This PR partially resolves #514.


andygrove (approved):
Thanks @zynaxsoft

Issue #581, opened 2021-02-20
Optimization: Read local shuffle files


andygrove:
In the following scenarios, shuffle files can be read directly from disk rather than streamed from an executor via the Flight server:

1. The shuffle was created by the current executor (i.e. an executor fetching from itself)
2. Multiple executors are running with a shared file system (such as when running a single node cluster)

PR #588, opened 2021-02-20
Shuffle reader stats

andygrove:
This is messy and I'll try and clean it up tomorrow. The goal is to make the statistics from execute_partition available to the shuffle reader that will read those results. This is a step towards the actual goal of choosing the smaller side of a join to be the build side.

This also implemented the local shuffle reader optimization.

Closes https://github.com/ballista-compute/ballista/issues/581


andygrove:
Replaced by https://github.com/ballista-compute/ballista/pull/636

Issue #567, opened 2021-02-18
Scheduler gRPC should support receiving SQL strings (as well as logical plans)


andygrove:
The scheduler gRPC method `ExecuteLogicalPlan` should be renamed to `ExecuteQuery` and the `ExecuteQueryParams` protobuf message should be updated to support receiving either a logical plan or a SQL string:

```protobuf
message ExecuteQueryParams {
  oneof query {
    LogicalPlanNode logical_plan = 1;
    string sql = 2;
  }
}
```

The implementation should be updated to support the SQL case and would simply create a logical plan from it and then delegate to the existing code for executing the logical plan. 

This is required by the JDBC driver and other clients that simply want to execute SQL against a cluster.

fitzoh:
Taking a closer look, is this made redundant by #577?

PR #590, opened 2021-02-21
ExecuteLogicalPlan  ->  ExecuteQuery  and allow plain SQL

fitzoh:
Changes `ExecuteLogicalPlan` to `ExecuteQuery` and allows plain SQL strings in addition to logical plans as described in #567.

New to both Rust and the project, any guidance on better/more idiomatic code very welcome.

### DIFF
--- a/rust/ballista/proto/ballista.proto
+++ b/rust/ballista/proto/ballista.proto
@@ -471,8 +471,10 @@ message RegisterExecutorParams {
 message RegisterExecutorResult {}
 
 message ExecuteQueryParams {
-  LogicalPlanNode logical_plan = 1;
-}
+  oneof query {
+    LogicalPlanNode logical_plan = 1;
+    string sql = 2;
+  }}
 
 message ExecuteSqlParams {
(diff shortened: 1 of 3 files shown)


andygrove:
Thanks @fitzoh this looks great - I can revert the change that I had made.

fitzoh:
I missed the comment about needing to getting a previously registered execution context, would probably want to pull that in right?

https://github.com/ballista-compute/ballista/blob/cf91fe8eb1dfe5bc20fa2d0f13e74c170d1ff23c/rust/ballista/src/scheduler/mod.rs#L179-L180

>Thanks @fitzoh this looks great - I can revert the change that I had made.

Appreciated, but don't do that on account of this PR if the other approach makes more sense as the a long term API.

andygrove (approved):
LGTM. Thanks @fitzoh

andygrove:
I filed https://github.com/ballista-compute/ballista/issues/593 for the follow up issue about the context

PR #608, opened 2021-02-27
Add option for standalone data path

djKooks:
Hello,
Looking on standalone mode, and seems there are options to make new standalone client by defining path or not.
(path for sled DB)

This is just a draft, so please confirm following feature is reasonable or not, and I will go on to the next step.

🙏


edrevo (commented):
Looks like a good start! I wonder if sled needs to be properly closed (I'm on a cell phone right now) in order to ensure that the file isn't corrupt. Right now we don't have any kind of graceful shutdown on the scheduler, so if sled needs to be flushed/closed, we would need this graceful shutdown for this change to be useful

djKooks:
@edrevo I've checked sled, and seems it does not require `close` handler after usage.

djKooks:
@edrevo how do you think of input name? Is `standalone_data_path` seems okay?

rust/executor/executor_config_spec.toml
```diff
@@ -59,4 +59,9 @@ abbr = "c"
 name = "concurrent_tasks"
 type = "usize"
 default = "4"
-doc = "Max concurrent tasks."
\ No newline at end of file
+doc = "Max concurrent tasks."
+
+[[param]]
+name = "standalone_data_path"
```
andygrove, line 65: Perhaps a better name could be `scheduler_data_path` or something similar?

djKooks: @andygrove seems good. But as I know this parameter will be only used under `standalone` mode. Will it not confuse?

djKooks:
@andygrove renamed~please check

andygrove:

PR #610, opened 2021-02-28
Bump Arrow version to f7cf157

andygrove:
There are some bug fixes we need to pick up

### DIFF
--- a/python/Cargo.toml
+++ b/python/Cargo.toml
@@ -23,8 +23,8 @@ tokio = { version = "1.0", features = ["full"] }
 serde_json ="1.0.61"
 serde = "1.0"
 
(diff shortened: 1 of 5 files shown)


edrevo (approved):

edrevo:
Looks like this change breaks the python bindings?

```
    Checking ballista-py v0.1.0 (/home/runner/work/ballista/ballista/python)
error[E0308]: mismatched types
  --> src/functions.rs:70:36
   |
70 |         expr: logical_plan::concat(expressions),
   |                                    ^^^^^^^^^^^ expected enum `datafusion::logical_plan::Expr`, found struct `std::vec::Vec`
   |
   = note: expected enum `datafusion::logical_plan::Expr`
            found struct `std::vec::Vec<datafusion::logical_plan::Expr>`
 ```

PR #617, opened 2021-03-01
fix #615

### LINKED ISSUE #615: Use PartitionId in TaskStatus
See

https://github.com/ballista-compute/ballista/blob/a0c609f3a43ed316df6c21c3e04ce7de726455d4/rust/ballista/proto/ballista.proto#L486-L496

vigsterkr:
use PartitionId in TaskStatus

### DIFF
--- a/rust/ballista/src/bin/executor.rs
+++ b/rust/ballista/src/bin/executor.rs
@@ -77,9 +77,7 @@ async fn poll_loop(
                     info!("Current task finished");
                     *running_task_guard = None;
                     task_status.push(TaskStatus {
-                        job_id: task_id.job_id,
-                        stage_id: task_id.stage_id,
-                        partition_id: task_id.partition_id,
+                        partition_id: Some(task_id),
                         status: Some(task_status::Status::Completed(CompletedTask {
                             executor_id: executor_meta.id.clone(),
                         })),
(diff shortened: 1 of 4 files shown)


rust/ballista/src/bin/executor.rs
```diff
@@ -77,9 +77,11 @@ async fn poll_loop(
...
                     *running_task_guard = None;
                     task_status.push(TaskStatus {
-                        job_id: task_id.job_id,
-                        stage_id: task_id.stage_id,
-                        partition_id: task_id.partition_id,
+                        partition_id: Some(PartitionId {
+                            job_id: task_id.job_id,
+                            stage_id: task_id.stage_id,
+                            partition_id: task_id.partition_id,
+                        }),
```
edrevo, line 84: can't you do this? (maybe you'll need to clone)

```suggestion
                        partition_id: Some(task_id),
```

rust/ballista/src/bin/executor.rs
```diff
@@ -90,9 +92,11 @@ async fn poll_loop(
...
```
edrevo, line 99: same comment, maybe you can just re-use task_id?

rust/ballista/src/bin/executor.rs
vigsterkr: 🤦 definitely

edrevo (approved):

edrevo:
Many thanks @vigsterkr

PR #619, opened 2021-03-02
Refactor cargo

edrevo:
I'm moving the code around to get real multi modules. This is just a first step to unblock the fat-LTO optimizations, but I think there's an opportunity to continue improving wrt physical code design.


rust/lib/Cargo.toml
```diff
@@ -15,12 +15,9 @@ simd = ["datafusion/simd"]
 snmalloc = ["snmalloc-rs"]
```
edrevo, line 15: @Dandandan: does is make sense to have a custom allocator inside a lib? I don't have much experience with allocators in rust, so I don't know what the best practice is here

docker/rust.dockerfile
```diff
@@ -1,43 +1,37 @@
...
-# Compile Ballista dependencies
-# the build script generates code based on the config specs
-# we make fake config specs here so we don't need to modify build.rs to build only our dependencies
-RUN mkdir -p /tmp/ballista/src/bin/ && echo 'fn main() {}' >> /tmp/ballista/src/bin/executor.rs && echo "" >> /tmp/ballista/src/bin/executor_config_spec.toml && echo "" >> /tmp/ballista/src/bin/scheduler_config_spec.toml
-RUN mkdir -p /tmp/ballista/proto
-COPY rust/ballista/proto/ballista.proto /tmp/ballista/proto/
-COPY rust/ballista/build.rs /tmp/ballista/
+FROM base as planner
+COPY rust .
+RUN cargo chef prepare --recipe-path recipe.json
```
edrevo, line 8: Cargo chef is a nice utility to improve cache hits when building on docker: https://github.com/LukeMathWalker/cargo-chef

edrevo:
@andygrove, ASL License check is failing due to `File at rust/lib/src/memory_stream.rs does not contain ASL license header`, but the file does have the ASL License. It does have a different text, though, because it came from datafusion. How do we handle this? Do we just ignore that check for this PR?

rust/lib/Cargo.toml
Dandandan: You are right, it should be included only in the binary project (mainly the executor as it will do the heavy work, in other places it probably doesn't hurt). That project will set the global allocator.
I think I included it in the lib as it also included the executor binary.

andygrove:
> @andygrove, ASL License check is failing due to `File at rust/lib/src/memory_stream.rs does not contain ASL license header`, but the file does have the ASL License. It does have a different text, though, because it came from datafusion. How do we handle this? Do we just ignore that check for this PR?

You could just add the file to the exclude list here: https://github.com/ballista-compute/ballista/blob/main/dev/ci/check-asl-header.py#L8-L11

andygrove (approved):
LGTM. Nice cleanup @edrevo

PR #620, opened 2021-03-03
Modify execute_action to stream results

jschmitz28:
**Description**
Modify execute_action to stream results instead of loading all results into a Vec. Not totally sure whether this is implemented as the TODO intended it to be so feel free to suggest any changes.

**Testing Done**
Ran TPCH queries as documented [https://github.com/ballista-compute/ballista/blob/main/rust/benchmarks/tpch/README.md](here) (1GB dataset) on my laptop with 1 scheduler + 3 executors and saw similar results in each (with high enough variance that maybe any real differences probably aren't detectable with my current setup).

I haven't gotten the docker setup to build on my laptop so I didn't run any of the docker based tests.

Before changes:
```
Query 1 avg time: 3472.13 ms
Query 3 avg time: 3435.16 ms
Query 5 avg time: 3431.22 ms
Query 6 avg time: 2901.66 ms
Query 10 avg time: 4013.10 ms
Query 12 avg time: 3720.07 ms
```

After changes:
```
Query 1 avg time: 3466.40 ms
Query 3 avg time: 3468.44 ms
Query 5 avg time: 3472.99 ms
Query 6 avg time: 2765.20 ms
Query 10 avg time: 4015.12 ms
Query 12 avg time: 3780.56 ms
```

### DIFF
--- a/rust/ballista/src/client.rs
+++ b/rust/ballista/src/client.rs
@@ -14,9 +14,12 @@
 
 //! Client API for sending requests to executors.
 
-use std::convert::{TryFrom, TryInto};
 use std::sync::Arc;
 use std::{collections::HashMap, pin::Pin};
+use std::{
+    convert::{TryFrom, TryInto},
+    task::{Context, Poll},
+};
 
 use crate::error::{ballista_error, BallistaError, Result};
 use crate::memory_stream::MemoryStream;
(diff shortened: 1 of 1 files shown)


andygrove (approved):
LGTM. Thanks @jschmitz28

Issue #592, opened 2021-02-21
Build binaries with LTO in Docker images


Dandandan:
Building with LTO improves (in memory) query performance something around ~10% in DataFusion, this will likely have a similar impact on Ballista.

To quote @edrevo https://github.com/ballista-compute/ballista/pull/558

> I think at some point we should move from the current cargo structure to a real multi-module structure: scheduler, executor, lib and benchmarks. That would allow enabling fat lto for the binaries.

PR #621, opened 2021-03-03
Build binaries with --release with lto by default

Dandandan:
https://github.com/ballista-compute/ballista/issues/592

Just putting it in the workspace works nicely.
Will do some comparison in benchmarking later, and maybe test with codegen units =1 too

### DIFF
--- a/docker/rust.dockerfile
+++ b/docker/rust.dockerfile
@@ -19,13 +19,10 @@ ARG RELEASE_FLAG=--release
 # force build.rs to run to generate configure_me code.
 RUN apt-get -y install cmake
 ENV FORCE_REBUILD='true'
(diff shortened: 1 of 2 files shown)


Dandandan:
@edrevo what do you think about this one? Should we build all release builds with lot or do we want more fine grained tuning? AFAIK the RUSTFLAGS approach doesn't yet work with the current structure.

edrevo:
I think it is fine to add lto to the release builds. I don't know when I would want a release non-lto build, to be honest.

Dandandan:
Let's then keep this. If release profile with lto is becoming a problem (LTO builds are a bit more than 2x as slow to build compared to non-lto) then we can look at using RUSTFLAGS

edrevo (approved):

PR #623, opened 2021-03-04
Executor control how many concurrent tasks run 

### LINKED ISSUE #580: Executor needs to control how many concurrent tasks run
The executor has a `concurrent_tasks` argument which is currently ignored. We need to control how many concurrent tasks can run per-executor in order to manage memory pressure.

boaz-codota:
Closes #580

### DIFF
--- a/rust/executor/src/execution_loop.rs
+++ b/rust/executor/src/execution_loop.rs
@@ -1,0 +1,149 @@
+use std::convert::TryInto;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::mpsc::{Receiver, Sender, TryRecvError};
+use std::{sync::Arc, time::Duration};
+
+use datafusion::physical_plan::ExecutionPlan;
+use log::{debug, error, info, warn};
+use tonic::transport::Channel;
+
+use ballista::serde::scheduler::ExecutorMeta;
+use ballista::{
+    client::BallistaClient,
+    serde::protobuf::{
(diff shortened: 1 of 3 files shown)


edrevo:
many thanks for this @boaz-codota! can you add the apache header to the new please?

boazberman:
Will do tomorrow. Thank you for reviewing!

rust/executor/src/execution_loop.rs
```diff
@@ -0,0 +1,149 @@
...
+        tokio::time::sleep(Duration::from_millis(250)).await;
+    }
+}
+
+async fn run_received_tasks(
+    mut executor_client: BallistaClient,
+    executor_id: String,
+    available_tasks_slots: Arc<AtomicUsize>,
+    task_status_sender: Sender<TaskStatus>,
+    result: tonic::Response<PollWorkResult>,
```
edrevo, line 71: Can't you borrow most of there parameters to avoid cloning every 250 millisecs?

boaz-codota: Nope, can't. Everything cloned is cloned to be moved to the `tokio::spawn` call

edrevo: but the tokio::spawn call only happens `if let Some(task) = result.into_inner().task`, which won't be true 99% of the time

boaz-codota:
@edrevo fixed

edrevo (approved):

PR #624, opened 2021-03-04
Fix dockerfile

edrevo:
The non-snmalloc binary was being built after the snmalloc binary, thus ending with a lower performance executor


edrevo:
i ran the integration tests manually and they passed

PR #627, opened 2021-03-05
Made hostname optional

### LINKED ISSUE #618: Make external hostname in executor optional
Having the ability to set an external/advertised hostname is great since it provides users a lot of flexibility in network deployments. However, having it a required argument is a pain for the most common scenario, where the scheduler, client and executors talk to each other in the same network (e.g. k8s or docker-compose).

We should make the external hostname optional. If the scheduler receives executor metadata without a hostname, it should register the caller's IP address as the hostname.

This will make it easier to deploy the executors as a kubernetes deployment, or to `docker-compose scale ballista-executor=<N>` in the integration tests.

### Code pointers

external hostname is defined here:
https://github.com/ballista-compute/ballista/blob/a0c609f3a43ed316df6c21c3e04ce7de726455d4/rust/ballista/src/bin/executor.rs#L172

executor metadata is defined here:
https://github.com/ballista-compute/ballista/blob/a0c609f3a43ed316df6c21c3e04ce7de726455d4/rust/ballista/proto/ballista.proto#L462-L466

scheduler saves executor metadata here:
https://github.com/ballista-compute/ballista/blob/a0c609f3a43ed316df6c21c3e04ce7de726455d4/rust/ballista/src/scheduler/mod.rs#L118-L125

msathis:
Fixes #618 

But not completely sure about the way ipv4 address is fetched.

### DIFF
--- a/rust/executor/src/main.rs
+++ b/rust/executor/src/main.rs
@@ -55,6 +55,23 @@ mod config {
 #[global_allocator]
 static ALLOC: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;
 
+pub fn get_system_ip() -> Option<String> {
+    let socket = match UdpSocket::bind("0.0.0.0:0") {
+        Ok(s) => s,
+        Err(_) => return None,
+    };
+
+    match socket.connect("8.8.8.8:80") {
+        Ok(()) => (),
+        Err(_) => return None,
+    };
+
+    return match socket.local_addr() {
+        Ok(addr) => Some(addr.ip().to_string()),
+        Err(_) => None,
+    };
+}
+
 #[tokio::main]
 async fn main() -> Result<()> {
     env_logger::init();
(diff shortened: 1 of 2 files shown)


rust/executor/src/main.rs
```diff
@@ -68,10 +86,12 @@ async fn main() -> Result<()> {
     }
 
     let namespace = opt.namespace;
-    let external_host = opt.external_host;
+    let external_host: String = opt.external_host.unwrap_or(get_system_ip().unwrap());
```
edrevo, line 89: I don't think you should be getting the system IP here. It will be tough to get this right from the executor's side, since the executor could have an arbitrarily complex network setup (multiple NICs, multiple IPs...).

I think it would be simpler if the registration information that the executor sends to the scheduler omits the hostname, and the scheduler could just use whatever IP address was used to talk to it (see https://docs.rs/tonic/0.4.0/tonic/struct.Request.html#method.remote_addr). Does that sound good?

edrevo (commented):
Many thanks for the PR!

rust/executor/src/main.rs
msathis: Yes. I understand it now. Will make the necessary modifications.

andygrove:

PR #630, opened 2021-03-05
Refactor modules

edrevo:
After thinking some more about physical design, here's a second, more thought out iteration. Code is now divided into 4 main crates:

- Core: contains the common interface code, like proto definitions, errors, etc.
- Executor: contains executor-specific code
- Scheduler: contains scheduler-specific code
- Client: contains the necessary code that a client would need to build and send an execution plan to the scheduler

The dependency graph is the following:
image

Executor depends on scheduler because we allow "standalone executors", which contain an in-proc scheduler. I'm thinking that after this refactor, it is probably worth removing that mode and including a "self-contained" feature in the client module that would spin up a scheduler+executor in-proc. That way, it would be truly easy to give Ballista a test run (no need to run any external process).

Another improvement would be to add features in the scheduler for etcd and standalone, to reduce the dependencies in certain scenarios.

I'll open issues for these improvements as soon as the PR is merged.


docker/rust.dockerfile
```diff
@@ -17,13 +18,8 @@ COPY --from=cacher /tmp/ballista/target target
 ARG RELEASE_FLAG=--release
 
 # force build.rs to run to generate configure_me code.
-RUN apt-get -y install cmake
 ENV FORCE_REBUILD='true'
 RUN cargo build $RELEASE_FLAG
-ARG EXECUTOR_BUILD_FEATURES='--features snmalloc'
```
edrevo, line 23: I'm enabling snmalloc as a default feature. I've figured that if we want to have the Docker images with snmalloc, then we should make that feature opt-out.

That means that cmake is not required to build Ballista

rust/client/src/columnar_batch.rs
```diff
@@ -13,7 +13,7 @@
 
 use std::{collections::HashMap, sync::Arc};
 
-use crate::error::{ballista_error, Result};
+use ballista_core::error::{ballista_error, Result};
```
edrevo, line 16: This code isn't referenced anywhere. Not sure if the plan is to use it at some point or if we should just delete it.

rust/executor/Cargo.toml
```diff
@@ -7,23 +7,28 @@ homepage = "https://github.com/ballista-compute/ballista"
 repository = "https://github.com/ballista-compute/ballista"
 authors = ["Andy Grove <andygrove73@gmail.com>"]
 edition = "2018"
-include = ["src/**/*", "Cargo.toml"]
```
edrevo, line 10: I have removed the include directives. I don't think they were actually doing anything useful. I am happy to revert if that isn't the case

andygrove:
This is a big PR to review so I haven't gone through this in any detail but I like the direction this is heading in.

andygrove (approved):

andygrove:
I'm going ahead with merging this since integration tests passed locally for me and I plan on rebasing and completing some pending PRs.

PR #636, opened 2021-03-07
Implement partition statistics

andygrove:
Changes in this PR:

- Add `PartitionStats` and `ColumnStats` to protobuf and implement serde code
- Distributed planner now stores partition stats from executed query stages
- Removes `null_count` that I had previously added in `PartitionStats`. This is a column-level statistic not a partition-level statistic so we should add it back once we implement column stats.

### DIFF
--- a/rust/core/proto/ballista.proto
+++ b/rust/core/proto/ballista.proto
@@ -461,8 +461,10 @@ message PartitionId {
 }
 
 message PartitionStats {
-  uint32 row_count = 1;
-  repeated ColumnStats column_stats = 2;
+  uint64 num_rows = 1;
+  uint64 num_batches = 2;
+  uint64 num_bytes = 3;
+  repeated ColumnStats column_stats = 4;
 }
 
 message ColumnStats {
--- a/rust/core/proto/ballista.proto
+++ b/rust/core/proto/ballista.proto
@@ -461,9 +461,9 @@ message PartitionId {
 }
 
 message PartitionStats {
-  uint64 num_rows = 1;
-  uint64 num_batches = 2;
-  uint64 num_bytes = 3;
+  int64 num_rows = 1;
+  int64 num_batches = 2;
+  int64 num_bytes = 3;
   repeated ColumnStats column_stats = 4;
 }
 
--- a/rust/core/proto/ballista.proto
+++ b/rust/core/proto/ballista.proto
@@ -450,6 +450,7 @@ message ExecutePartition {
 message PartitionLocation {
   PartitionId partition_id = 1;
   ExecutorMetadata executor_meta = 2;
+  PartitionStats partition_stats = 3;
 }
 
 // Unique identifier for a materialized partition of data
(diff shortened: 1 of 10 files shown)


andygrove:
@Dandandan fyi

rust/core/src/serde/scheduler/mod.rs
```diff
@@ -87,6 +90,99 @@ impl From<protobuf::ExecutorMetadata> for ExecutorMeta {
     }
 }
 
+/// Summary of executed partition
+#[derive(Debug, Copy, Clone)]
+pub struct PartitionStats {
+    num_rows: u64,
+    num_batches: u64,
+    num_bytes: u64,
```
Dandandan, line 98: Are the statistics always known? How is the case encoded when the statistics are not available?

rust/core/proto/ballista.proto
```diff
@@ -459,6 +460,20 @@ message PartitionId {
...
+  uint64 num_batches = 2;
+  uint64 num_bytes = 3;
+  repeated ColumnStats column_stats = 4;
+}
+
+message ColumnStats {
+  ScalarValue min_value = 1;
+  ScalarValue max_value = 2;
+  uint32 null_count = 3;
+  uint32 distinct_count = 4;
```
edrevo, line 474: Isn't this going to be expensive to compute? Or is this projected as an `Option<i32>` and it is expected to be optional?

Dandandan: I think all statistics should be optional. And yes distinct is expensive to compute / we can do a lot with min/max+row count already.

andygrove: I agree that statistics should always be optional. Also, statistics should always be assumed to be estimates. There are times when distinct is cheap to compute (for the query `SELECT foo, COUNT(*) GROUP BY foo ORDER by foo DESC LIMIT 100` we can assume that `distinct_count <= 100`).

rust/core/src/serde/scheduler/mod.rs
andygrove: I've updated to use `Option<u64>` with a `-1` value in the protobuf representing `None`

andygrove:
Integration tests pass locally now

rust/core/src/utils.rs
```diff
@@ -169,20 +70,17 @@ pub async fn write_stream_to_disk(
             .iter()
             .map(|array| array.get_array_memory_size())
             .sum();
-        let batch_null_count: usize = batch.columns().iter().map(|array| array.null_count()).sum();
         num_batches += 1;
```
Dandandan, line 73: 👍

rust/scheduler/src/planner.rs
```diff
@@ -325,21 +314,32 @@ async fn execute_query_stage(
...
             let mut client =
                 BallistaClient::try_new(&executor_meta.host, executor_meta.port).await?;
-            client
-                .execute_partition(job_id, stage_id, partition_ids, plan)
-                .await
+            let stats = client
+                .execute_partition(job_id.clone(), stage_id, partition_ids.clone(), plan)
+                .await?;
+            let mut meta: Vec<PartitionLocation> = Vec::with_capacity(partition_ids.len());
+            for part in &partition_ids {
```
Dandandan, line 321: Could be using iterator here

Dandandan (approved):
Nice, LGTM (with lint error fixed) - also one additional style suggestion

PR #644, opened 2021-03-09
Implement serde for RepartitionExec physical operator

andygrove:
I'm working on implementing shuffle hash join and found that this serde code was missing,

### DIFF
--- a/rust/core/proto/ballista.proto
+++ b/rust/core/proto/ballista.proto
@@ -224,9 +224,9 @@ message RepartitionNode{
   }
 }
 
-message HashRepartition{
+message HashRepartition {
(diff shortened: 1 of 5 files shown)


andygrove:
@Dandandan fyi

rust/core/proto/ballista.proto
```diff
@@ -224,9 +224,9 @@ message RepartitionNode{
   }
 }
 
-message HashRepartition{
+message HashRepartition {
   repeated LogicalExprNode hash_expr = 1;
-  uint64 batch_size = 2;
+  uint64 partition_count = 2;
```
Dandandan, line 229: :+1:

PR #648, opened 2021-03-14
Api Server for Scheduler

msathis:
~~Not compiling yet. Struggling with borrow checker to fix the lifetime issues~~


boazberman:
I actually also started working on this, but I'll gladly leave the stage to you.

I'd just point out that I think that we should try to use a single tower abstraction and reuse the same port for this.

Tonic has an example on how to do this: https://github.com/hyperium/tonic/blob/master/examples/src/hyper_warp_multiplex/server.rs

Just to improve their example, we can parse the Request Type header to match for grpc, instead of sacrificing http2 to be grpc only.

Let me know if there is anything I can do to assist.

msathis:
Thanks @boazberman !! That was a great suggestion. I've updated the current PR with your idea. Please review it.

msathis:
Screenshot 2021-03-19 at 4 07 23 PM

boazberman:
I'm not in a position to review PRs in this repo, but I think you used the wrong header- Accept is an header used to indicate what type of response the client would like to receive, not what kind of request he sent.
For that there is Content Type: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

msathis:
@boazberman I think this is the right way to do content negotiation. Server will read what type of resource client wants & sends it accordingly. https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation#server-driven_content_negotiation

msathis:
@edrevo @andygrove Review please :)

andygrove:
Thanks @msathis I'm excited to try this out. I hope to have time later today to test it and review the PR. Note that we are in the process of donating this project to Apache Arrow so potentially we may have to create this PR over there, building on https://github.com/apache/arrow/pull/9723

msathis:
@andygrove Sure! That wont be a problem :)

andygrove:

andygrove:
@msathis do you want to re-open this against the Arrow repo? https://github.com/apache/arrow

